PR #35, opened 2015-02-19
createSupportedKeySystem() should return a Promise

dmlap:
Add missing return statement to the helper function in Example 4


ddorwin:
Thanks for finding and fixing this.

index.html is generated. Instead, you need to modify encrypted-media-respec.html, which is used to generate index.html. (I can do the generation next time I commit a change.)

dmlap:
Ah, I missed the template. Re-opening as #37.

Issue #22, opened 2015-01-24
Request: Allow CDMs to support and enforce resolution restrictions (and expose the status of those restrictions to the application)


ddorwin:
\[**March 6, 2015 Update**: The discussion has diverged from the original intent of this bug, including the text of this comment, and the first commit below. I have updated the title to reflect the discussion\]

[Bug 25409](https://www.w3.org/Bugs/Public/show_bug.cgi?id=25409) ("Allow applications to detect whether a key is usable before using it to decrypt content") explains the original motivation for the `keyStatuses` attribute and [Bug 25092 comment 33](https://www.w3.org/Bugs/Public/show_bug.cgi?id=25092#c33) explains the intended behavior and usage. However, this may not be clear from the spec text.

mwatson2:
In thinking about this further, I realize some of my previous comments may have been unclear / missing some points. Apologies for that. The following takes precedence if there is conflict ;-)

First, it's obviously desirable to be able to detect before playback starts what the key status is - particularly if it is "not allowed". It's highly desirable that CDMs attempt to engage HDCP *before* playback starts, since the UX for engaging HDCP during playback is horrible. I would expect implementations to attempt to engage HDCP as soon as the license is received and to indicate the key status once that process is complete. In that case the key status will remain fixed from then on. [If there are implementations which wait to receive content before engaging HDCP, then the key status could change, but that is a horrible UX as noted].

Second, we do have a requirement for content-specific policies associated with a single key. Whilst EME CDMs might support separate keys with separate content-independent policies, many services are prevented from using this capability by the need to have a single set of streams shared with many other (non-EME) devices which do not support this.

"Downscaling" is an example of a policy which is always content-dependent. If the policy is "downscale to SD", then this is obviously a noop if the content is SD resolution or lower. If the policy has exactly two tiers, then our present solution - a single key with status "output-downscaled" - is good: the application knows the tiers and therefore knows what content will pass the policy unmodified. How to handle >2 tiers is an open issue. (For example, with UHD the HDCP version matters as well and there is a "downscale to HD" case) 

"Not allowed" is not so cleanly supported presently. As above, if the policy has two tiers, then the "output-not-allowed" value could be used for the case of HD not allowed but SD allowed. The application knows the tiers and can distinguish from the case where nothing is allowed since that case would result in an error. So it can assume that SD is all that is allowed. A better name would be "output-restricted".

[...]

PR #65, opened 2015-06-16
Issue #22: Rename output-not-allowed MediaKeyStatus to output-restricted

mwatson2:
Proposal to rename `MediaKeyStatus` `output-not-allowed` to `output-restricted` to address issue 22. The existing `output-downscaled` status should be used only if content really will be downscaled and so is modified to reflect that. `output-restricted` covers the cases where output is not allowed for any content or restricted to content of certain resolutions etc.


encrypted-media-respec.html
```diff
@@ -1498,17 +1498,19 @@
             The time represented by the <a def-id="expiration"></a> attribute MUST be earlier than the current time.
             All other keys in the session MUST have this status.
           </dd>
+          <dt>output-restricted</dt>
+          <dd>
+            There are output restrictions associated with the key which cannot currently be met.
```
ddorwin: The descriptions for the two output-* values should be as close as possible. I suggest copying these (potentially slightly modified) to output-downscaled below. Probably with line 1510 in between them.

encrypted-media-respec.html
```diff
@@ -1498,17 +1498,19 @@
...
+            The application should avoid using streams which will trigger the output restrictions associated
```
ddorwin: nit: s/which/that/

encrypted-media-respec.html
```diff
@@ -1498,17 +1498,19 @@
...
+            with the key.
```
ddorwin: We should specify what will happen. We could either add that to the end of this sentence (i.e. "... to avoid being unable to encrypt it") or in between these two sentences like I have proposed for "output-downscaled" in my first comment above.

encrypted-media-respec.html
```diff
@@ -1498,17 +1498,19 @@
...
+          </dd>
           <dt>output-downscaled</dt>
           <dd>
             The key is not currently usable to decrypt <a def-id="media-data"></a> at full quality (e.g. resolution) because its output requirements cannot currently be met.
-            Media data decrypted with this key may be presented at a lower quality (e.g. resolution).<br/>
+            Media data decrypted with this key will be presented at a lower quality (e.g. resolution).<br/>
```
ddorwin: nit: "will" seems more absolute than we mean. It will only be lower if necessary to meet the output requirements. Perhaps, "will... if necessary to meet the output requirements."

encrypted-media-respec.html
```diff
@@ -1498,17 +1498,24 @@
             The time represented by the <a def-id="expiration"></a> attribute MUST be earlier than the current time.
             All other keys in the session MUST have this status.
           </dd>
+          <dt>output-restricted</dt>
+          <dd>
+            There are output restrictions associated with the key that cannot currently be met.
+            Media data decrypted with this key may be blocked from presentation, if necessary according to
+            the output restrictions.
+            The application should avoid using streams which will trigger the output restrictions associated
```
ddorwin: nit: s/which/that/ here and at 1514.

ddorwin:
One nit. Also, the description should be updated to reflect the rename rather than addition. Other than that, LGTM. Thanks.

Issue #63, opened 2015-06-01
Align normative text on user consent for Distinctive Identifiers with privacy section


mwatson2:
The non-normative Privacy sections states (Section 11.4.2):

"User agents must ensure that users are fully informed and/or give explicit consent before Distinctive Identifier(s) are exposed, such as in messages from the Key System implementation."

However, the option to provide information has been dropped in the normative procedures (Section 3.1.2.1, Step 15):

"If there is no persisted consent covering accumulated configuration for the origin, it is recommended that implementations request user consent to use Distinctive Identifier(s)."

The Privacy text has been stable for some time and we have not discussed recommending consent in all cases. Whether explicit consent is necessary is a matter for user agents.

I propose modifying the text in 3.1.2.1 to the following:

"If explicit user consent for the use of Distinctive Identifiers is required and if there is no persisted consent covering accumulated configuration for the origin, it is recommended that implementations request user consent to use Distinctive Identifier(s)."

jdsmith3000:
I agree with Mark's analysis, and his proposed change.  I believe we've not previously reached a decision that user consent is required to use distinctive identifiers, and the current spec language seems to suggest.

ddorwin:
This text was added six months ago to address [bug 27165](https://www.w3.org/Bugs/Public/show_bug.cgi?id=27165) from the TAG.

mwatson2:
The TAG bug referred to the case of unclearable identifiers. We have done better than the TAG asked by making those non-compliant. This issue is not about that.

This issue is about the case of clearable (site-specific) identifiers, where the Privacy section has always required either that users are fully informed or that they give consent.

However, the normative procedures refer only to the case of explicit consent.

ddorwin:
The bug summary refers to "unclearable identifiers", but the description refers to "identifiers that cannot be cleared along with regular cookies and site data." The [spec text](https://w3c.github.io/encrypted-media/#allow-identifiers-cleared) related to "cookie-like" clearing uses `SHOULD` and `RECOMMENDED`, so there would appear to be scenarios where the intent of the bug is not met (even with the current text _recommending_ obtaining user consent).

If you want to make a proposal tying all of these together, that's fine. However, my reading of your current proposal is that it would *recommend* requesting consent if consent is *required*. That doesn't make sense - it's not actually required when required. You could drop `RECOMMENDED`, making it required when required, but you'd need to define when it is required.

[...]

PR #66, opened 2015-06-16
Issue #63: Clarify the conditions in which consent for distinctive id…

mwatson2:
Clarify the conditions in which consent for distinctive identifiers is required.

Consent is required if the recommendations for clearing identifiers together with cookies etc. are not supported or if the user agent decides for other reasons that consent is required. Otherwise informing the user is a possible option per TAG guidance.


mwatson2:
@ddorwin I have tried to address your comments.

I have two questions, though:
(1) We do not describe the persisting of user consent status. Should we, or should it be left to User Agents do decide whether and when to persist the consent status (perhaps based on a "do not ask me again" checkbox?).
(2) We do not have any "Denied" consent status, only ConsentRequired, InformUser and Allowed. That is, we can't persist the "Denied" status. Should UAs be allowed to persist a "Denied" status or are they required always to ask again ?

ddorwin:
It looks like the rebase or some other action removed my comments from this PR and made them unreachable. For the record, they are at https://github.com/mwatson2/encrypted-media/commit/0d33734864311d6c61ad0f1af3f2213fd0611e39.

ddorwin:
Replies to @mwatson2's questions:
(1) Good point. I think it's up to the UAs under what conditions to persist. However, it probably makes sense to mention that this MAY happen after the request. Thus, something simple like:
> The user agent MAY choose to persist the user response.

(2) Another good point. I believe UAs do persist denials for permissions. Chrome does, at least. The algorithms should allow that, though it is not required for UAs to implement it. I think we can address this by duplicating step 1 of Get Distinctive Identifier Consent Status and changing the new step 1 to "persisted denial" or "persisted blocking" and returning Denied.

Another minor item:
(3) I think we should add something like "and wait for the user response" to the end of the "Request user consent" step.

Other than the two comments above and these issues, this LG. Thanks!

mwatson2:
Hmm, that would probably be the rebase, as it required a push --force to put up the new commits. I guess the alternative is not to rebase but to merge and then only do the rebase before finally merging the pull request ?

ddorwin:
Hmm, probably. I guess we'll learn how to use the tool eventually. In the meantime, it's probably a good practice to save the commit URL before doing any forced push in case we need to manually post a comment pointing to it.

mwatson2:
I believe this is ready to merge now. Any further comments from anyone ?

ddorwin:
A few more minor comments. Also, see my first two comments above from yesterday. Each is for a separate commit. (GitHub doesn't appear to allow commenting on lines that weren't modified by a commit.)

mwatson2:
Merged as 081f5696bd8315745d3bca29e92712613219e895

Issue #80, opened 2015-08-27
"tracked" sessions: record of usage should track session usage, not individual key usage


ddorwin:
The text from pull request #54 says "the record of key usage" contains an "entry for block key ID." This means that there is a record of usage for each key in the session. This granularity seems unnecessary and has costs, including:

1. Complicating the CDM, license protocol, and license server implementation
2. Requiring the author's server to aggregate the intervals from all keys to come up with the total playback period for the session

The spec generally treats sessions as a single entity, including closing, removing, and expiration time, and it's unclear why this should be different. If an author really needs different usage data, he or she can use multiple sessions.

mwatson2:
Agreed. We should record the overall start and end times and the ids of the keys that were available.

jdsmith3000:
Tracking by session rather than individual keys aligns with Microsoft's thinking.  We support this recommendation.

paulbrucecotton:
If we have initial agreement here, can one of editors volunteer to prepare a pull request for this issue?

mwatson2:
Yes, I will do that. So for the others where there is agreement or at least
a direction to discuss in more detail.

...Mark

Sent from my iPhone

On Sep 3, 2015, at 6:55 AM, Paul Cotton <notifications@github.com> wrote:

If we have initial agreement here, can one of editors volunteer to prepare
a pull request for this issue?

—
Reply to this email directly or view it on GitHub
<https://github.com/w3c/encrypted-media/issues/80#issuecomment-137455870>.

mwatson2:
PR: https://github.com/w3c/encrypted-media/pull/87

Issue #59, opened 2015-05-21
|expiration| should be Unix time like Date()


ddorwin:
The `expiration` attribute is currently defined as
>The time, in milliseconds since 01 January, 1970 UTC, ...

This could be interpreted as including leap seconds, which would appear to be inconsistent with ECMAScript and `Date()`, which [appear to expect Unix time](https://people.mozilla.org/~jorendorff/es6-draft.html#sec-time-values-and-time-range).

Ideally, there would be a definition of time that all specs could point to. I'll follow up with Web IDL.

ddorwin:
We now have several instances of "time" in the spec. Even if we can't get a common definition in Web IDL, we should have a local definition of time in the Definitions section that those instances can refer to.

Once I follow up with Web IDL, we can mark this "to be implemented". The proposal is:
* Add a definition of "Time" to https://w3c.github.io/encrypted-media/#definitions.
* Include the current text about 1970 UTC.
* Further clarify that this is Unix time per the original comment.

ddorwin:
Also, define what an invalid time is. In issue #58, we decided on `NaN`. We also use `NaN` as the default value in algorithms.

ddorwin:
`Date` is still in the Web IDL spec and the bug to remove it [Web IDL bug 22824](https://www.w3.org/Bugs/Public/show_bug.cgi?id=22824) has not been updated in two years. I requested an update there.

For future reference, we removed the use of `Date` from EME for [bug 25902](https://www.w3.org/Bugs/Public/show_bug.cgi?id=25902).

ddorwin:
Note that https://heycam.github.io/webidl/#idl-Date already has the text about ECMAScript consistency.

If we do need a definition of time, perhaps the _Common definitions_ section of the Web IDL spec would be an appropriate place.

bzbarsky:
Yes, a type like `DOMTimeStamp` but explicitly consistent with ES `Date` timestamps would be a good idea.  And then auditing existing `DOMTimeStamp` users....

travisleithead:
Looks like the DOM standard is also expecting a DOMTimeStamp to be defined (```Event.timestamp```). Right now it's a broken link :(

paulbrucecotton:
@travisleithead or @plehegar  - Can you provide an update on the WebIDL issue 2 and recommend how we should proceed here?  Do we wait for Issue-2 to be processed?

mwatson2:
[...]

Issue #58, opened 2015-05-21
Clarify that the value of |expiration| should be "NaN" for licenses that never expire


ddorwin:
The definition of the `expiration` attribute currently says:
>The time, in milliseconds since 01 January, 1970 UTC, after which the key(s) in the session will no longer be usable to decrypt media data, or `NaN` if no such time exists, as determined by the CDM.

We should clarify that "no such time exists" includes licenses that (explicitly) never expire. Also, `Infinity` should never be used.

Note: `NaN` can be difficult to deal with in JavaScript, but I still think it is the best option. `Date(0)` and `Date(null)` both result in the Unix Epoch. `Date(undefined)` results in an invalid time like `Date(NaN)`, but I don't think Web IDL allows attributes to have the value `undefined`.

PR #87, opened 2015-09-03
Issue #80: Times in the record of key usage should be at session level

mwatson2:


mwatson2:
I agree with all @ddorwin 's suggestions.

For the questions:

(1) Do you need to add the key IDs here or is that already handled by update()?

They need to be added somewhere. I'll check and address.

(2) Is "tfirst" consistent with good JSON naming practices?

I don't know. Happy to take suggestions.

ddorwin:
The needs implementation label was making searches confusing, so I removed it.

encrypted-media-respec.html
```diff
@@ -1011,11 +1011,30 @@
...
+              </p>
+            </li>
+            <li>
+              <p>
+                The <dfn id="latest-decryption-time">latest decryption time</dfn>, defined as the latest time at which the session was used to decrypt content.
+              </p>
+            </li>
+          </ul>
+          <p>
+              A <a def-id="message"></a> of type <a def-id="message-type-license-release"></a> containing the <a def-id="record-of-key-usage"></a> will be generated when <a def-id="remove"></a> is called until the record is acknowledged by a response passed to <a def-id="update"></a>.
```
ddorwin: We should probably look at the behavior of sessions and calling remove() after the record is acknowledged in an update() call. Nothing specific to change in this PR, but this text makes me wonder.

encrypted-media-respec.html
```diff
@@ -1240,7 +1259,31 @@
...
                         <dt><a def-id="persistent-usage-record-session"></a></dt>
                         <dd>
-                          <p>Let <var>requested license type</var> be a non-persistable license that will persist a <var>record of key usage</var>.</p>
+                          <ol>
+                            <li>
+                              <p>Initialize this object's <var>record of key usage</var> as follows.</p>
+                              <ul>
+                                <li>
+                                  <p>
+                                    Set the list of <a def-id="key-id">key IDs</a> <a href="#known-key">known</a> to the session to the empty list.
```
ddorwin: nit: I'm not sure "the empty list" is a thing like "the empty string." Use "an"?

encrypted-media-respec.html
```diff
@@ -1240,7 +1259,31 @@
...
+                                  </p>
+                                </li>
+                                <li>
+                                  <p>
+                                    Set the <var>first decrypt time</var> to <code>null</code>.
```
ddorwin: I think we should use `NaN`. (I updated #59 to address this as well.)

ddorwin: Link to the "first-decryption-time" definition?

encrypted-media-respec.html
```diff
@@ -1011,11 +1011,30 @@
...
+            A session for which the license and any key(s) it contains SHALL NOT be persisted and for which a <a def-id="record-of-key-usage"></a> SHALL be persisted.
+            The <dfn id="record-of-key-usage">record of key usage</dfn> consists of:
+          </p>
+          <ul>
+            <li>
+              <p>A record of the <a def-id="key-id">key IDs</a> of all the key(s) that were at any time <a href="#known-key">known</a> to the session,</p>
+            </li>
+            <li>
+              <p>
+                The <dfn id="first-decryption-time">first decryption time</dfn>, defined as the time at which the session was first use to decrypt content, and
```
ddorwin: This is basically defining a fake (i.e. not exposed to the application) member variable like we do for various state of MediaKeys and MediaKeySession. In those cases, we just use <var>.

encrypted-media-respec.html
```diff
@@ -2138,12 +2194,12 @@
                                 <ol>
                                   <li>
                                     <p>
-                                      If the <var>record of key usage</var> does not contain an entry for <var>block key ID</var>, add a new entry for <var>block key ID</var> with <var>first decrypt time</var> set to the current time.
+                                      If the <var>first decrypt time</var> of the object's <var>record of key usage</var> is <code>null</code>, set the <var>first decrypt time</var> to the current time.
```
ddorwin: `NaN`

encrypted-media-respec.html
```diff
@@ -2138,12 +2194,12 @@
...
                                     </p>
                                   </li>
                                   <li>
                                     <p>
-                                      Set the <var>latest decrypt time</var> of the entry for <var>block key ID</var> in the <var>record of key use</var> to the current time.
+                                      Set the <var>latest decrypt time</var> of the object's <var>record of key usage</var> to the current time.
```
ddorwin: nit: You might be able to simplify this using a local variable named _usage_. Then you can set _usage_'s <var>latest decrypt time</var>. Similar in the previous step.

encrypted-media-respec.html
```diff
@@ -2551,18 +2607,12 @@
...
-                <dt>"tfirst"</dt>
-                <dd>The <var>first decrypt time</var> for this key ID expressed as a number giving the time, in milliseconds since 01 January, 1970 UTC.</dd>
-                <dt>"tlatest"</dt>
-                <dd>The <var>latest decrypt time</var> for this key ID expressed as a number giving the time, in milliseconds since 01 January, 1970 UTC.</dd>
-              </dl>
-            </dd>
+            <dt>"kids"</dt>
+            <dd>An array of <a def-id="key-id">key IDs</a>. Each element of the array is the base64url encoding of the octet sequence containing the key ID value.</dd>
+            <dt>"firstTime"</dt>
+            <dd>The <a def-id="first-decrypt-time"></a> expressed as a number giving the time, in milliseconds since 01 January, 1970 UTC.</dd>
```
ddorwin: This is the only place the definition is used. That's sort of useful, but it doesn't appear we do that in other items in this section, including "kids".

ddorwin: Everywhere else uses "first-decrypt**ion**-time".

encrypted-media.js
```diff
@@ -159,6 +159,9 @@
     'session-id': { func: term_helper, fragment: 'session-id', link_text: 'Session ID'  },
     'distinctive-identifier': { func: term_helper, fragment: 'distinctive-identifier', link_text: 'Distinctive Identifier'  },
     'distinctive-identifiers': { func: term_helper, fragment: 'distinctive-identifier', link_text: 'Distinctive Identifiers'  },
+    'record-of-key-usage': { func: term_helper, fragment: 'record-of-key-usage', link_text: 'record of key usage'  },
+    'first-decryption-time': { func: term_helper, fragment: 'first-decryption-time', link_text: 'first decryption time'  },
```
ddorwin: We probably don't need these two per earlier comments.

encrypted-media-respec.html
mwatson2: Well, certainly mathematically speaking there is only one empty list just as there is only one empty set, only one number zero etc. I'll admit that in Javascript one can have many distinct objects where the *value* of each one is the empty list. The "list of key ids known to the session" is an abstract (unspecified) kind of list, rather than a concrete type such as a Javascript Array object, so I think what is written would be ok, but - since we are in abstract type land - "an" would be ok too and may be potentially less confusing.

encrypted-media-respec.html
mwatson2: *record of key usage*

encrypted-media.js
mwatson2: I'm not sure I understand: it was your earlier comment which caused me to use <dfn> for these two and that seemed to necessitate adding these definitions to the JS in order to get links to work.

ddorwin: I meant my previous comments above, specifically those at lines 1023 and 2613. I was just saying that if we remove the `<dfn>`, we won't need these references.

encrypted-media-respec.html
mwatson2: I think this is a slightly different usage from `expiration` where we decided in `NaN` in issue #58. In that case, this is a real value for `expiration` with the meaning that there is no expiration. In this case we are using `null` as a temporary placeholder to indicate that the first / latest decryption has not happened yet. It's not much of a difference but it's not obvious that there's any particular reason to align on `NaN`.

encrypted-media-respec.html
mwatson2: Do you mean we should use `<var>` for first decrypt time instead of `<dfn>` ? That's fine for me, but I thought I introduced the `<dfn>` in response to one of your earlier comments ?

encrypted-media-respec.html
ddorwin: Sounds good. As you say, this is internal state. (We might still change this later depending on the outcome of #59. Regardless, it doesn't impact the visible behavior...

Except, for Clear Key. Do we then need to handle the conversion from "null" to some value in the JSON? Would it be simpler to just use `NaN` here then?

encrypted-media-respec.html
ddorwin: Yes, that is what I meant here. Yes, I suggested `<dfn>`, but when I looked at other examples, that didn't seem appropriate. We use `<dfn>` for terms like **Initialization Data Type** but not for internal/local variables like this.

I think it would be fine to define "First Decryption Time" if you want to reference the term elsewhere, but I believe the internal variable should be a `<var>`.

You could probably handle both here by saying something like the following. It's not exactly the best structure, but it works.
```<var>first decryption time</var> - The <dfn id="first-decryption-time">first decryption time</dfn>, which is the time at which the session was first use to decrypt content.```

encrypted-media-respec.html
mwatson2: JSON can't represent `NaN`, but it can represent `null` ;-)

mwatson2:
Are we ready to merge this now ? (No green button will be pressed ...)

ddorwin:
Oh, I missed that there were new commits as I only saw the comments in email. I will try to take a look tomorrow.

ddorwin:
The last two (non-merge) commits LGTM. Thanks.

mwatson2:
Merged as 7a55532a1ee19f937451e4e366d4b27c5f01099b

Issue #52, opened 2015-04-28
Remove reference to keys in Initialization Data definition


steelejoe:
Specifically this sentence in the Initialization Data [Definition](https://w3c.github.io/encrypted-media/#definitions):
It MUST NOT contain application data, client-specific data, user-specific data, **key(s)**, or executable code.

Should become this sentence:
It MUST NOT contain application data, client-specific data, user-specific data, or executable code.

steelejoe:
I forgot to add this reference. DASH-IF-OIP-v3.0 section 7.2 paragraph 5 has a specific reference to PSSH including keys.

ddorwin:
Per the telecon, this would be resolved at the same time as (retitled) #41.

jdsmith3000:
The ISO spec on Common Encryption is ISO_IEC_23001-7_2015(E).  It includes language in paragraph 9.1 that specifically says keys may be included in initiData.  We should not include conflicting requirements in the EME specification.

I see that this issue is listed as blocked on either issue #41 or issue #53.  I don't believe that is correct.  We should remove "keys" now from items prohibited in the initiData, and work these other two issues going forward.

steelejoe:
Extracting from his last email on this subject -- David proposed this text: 
"However, in order to make the text more technically sound, I propose we remove "key(s)" from the current sentence and add: Implementations MUST NOT use Initialization Data for anything other than generating a license request. Specifically, Initialization Data or portions of it MUST NOT be persisted, used to configure the CDM or session, or used directly by the implementation to extract or derive key(s)."
We can append "unless explicitly stated in the algorithms" as appropriate when adding support for use cases (i.e. Issues 41 and 53) that need more flexibility.

My counter proposal is that we simply remove "key(s)" from the sentence as in my initial proposal -- making the new definition: 
"It MUST NOT contain application data, client-specific data, user-specific data, or executable code."

The spec already covers what can be persisted as part of a session. Any text restricting how the potentially DRM-specific initData is used by the CDM beyond its storage is outside of the scope of this spec.

mwatson2:
+1 to Joe's last comment.

steelejoe:
Adding a tie-in to the ongoing email thread -- 
https://lists.w3.org/Archives/Public/public-html-media/2015Sep/0003.html

ddorwin:
[...]

PR #88, opened 2015-09-03
Issue #52: Remove restrictions on initData

mwatson2:
- Remove restrictions on initData
- Enforce no use by CDM of initData directly from media


ddorwin:
#52 has been addressed with PR #92. This can probably be closed.

paulbrucecotton:
@mwatson2:  Can we discard this PR since it is no longer needed?

mwatson2:
Addressed in #92.

Issue #82, opened 2015-08-27
"tracked" sessions: when to persist "record of key usage" is not specified


ddorwin:
Pull request #54 added text to section 7.5.2, the [Encrypted Block Encountered](https://w3c.github.io/encrypted-media/#encrypted-block-encountered) algorithm, that specifies modification of an entry of the "record of key usage". However, that record and the behavior around it is not specified. As has been [previously discussed](https://lists.w3.org/Archives/Public/public-html-media/2015Jun/0021.html), there are specific implementation requirements in order for such records to be useful to the author. These should be specified.

The method that is technically possible for most user agents involves storing such records in (tamper-resistant) memory then persisting the record when the license/keys are destroyed.

Therefore:
* The steps in 7.5.2 should have a Note to be clear that the data should not be persisted at that point.
* The algorithms related to license/key destruction should be updated to include persisting the "record of key usage." I believe this includes the Session Close algorithm and the remove() algorithm.

mwatson2:
Requirements on exactly how this is implemented arise from the fact that CDMs are subject to robustness rules. This applies to other aspects of EME as well but we have not so far discussed robustness in our specification. What should be the rule for deciding which robustness aspects we need to address in the specification and which we do not ?

ddorwin:
From previous discussions, I think we agree that:
* The usage records must be persisted to be effective in handling some cases, such as non-graceful close.
* Without certain platform capabilities, persisting the records at the wrong time leads to an unusable implementation.

This issue is requesting definition of the observable behavior in the algorithms as well as any additional information (not covered in the algorithms) necessary to ensure an implementation is interoperable and usable. This does not prevent implementations from using platform capabilities to enable greater protection (“robustness”) of the storage or optimizations, as long as the observable behavior is consistent with the algorithms.

mwatson2:
I agree the observable behavior should be clear and I'll review the current text in the light of the points above (next week).

However, I'm still stuck on how we should decide which robustness aspects to address and which not. For example, with persistent sessions, CDMs can support features such as play counts or license expiration (e.g. 24 hour rental) for offline playback which typically, in practice, require certain robustness features of the persistent store and clocks, but we do not address this at all in our specification. The presence or absence of the robustness features does not affect the behavior on the API. The level of robustness required is up to service providers. Should we address all this too ?

mwatson2:
[...]

PR #89, opened 2015-09-03
Issue #82: Clarify persistence of record of key destruction

mwatson2:
Also note implementation options for user agents.


ddorwin:
Let's finish and merge #87 first since this is touching some of the same text.

mwatson2:
I have updated this PR following the merge of PR#87. Comments please!

encrypted-media-respec.html
```diff
@@ -1027,6 +1027,10 @@
           <p class="note">
             Because the license and keys are not persisted, this record implicitly proves that the keys are no longer available in the session.
           </p>
+          <p class="note">
+            User agents MAY implememt this mechanism by means other than persisting data on key destruction - for example by persisting data during playback which is later used
+            to infer the fact of key destruction - provided the API behavior is compliant to this specification.
```
ddorwin: nit: s/API/observable/ ?

encrypted-media-respec.html
```diff
@@ -1876,8 +1886,18 @@
...
+            <li>
+              <p>
+                Destroy any license and/or keys associated with <var>session</var> that have not been explicitly stored.
+              </p>
+            </li>
             <li><p>Run the <a def-id="update-key-statuses-algorithm"></a> algorithm on the <var>session</var>, providing an empty sequence.</p></li>
             <li><p>Run the <a def-id="update-expiration-algorithm"></a> algorithm on the <var>session</var>, providing <code>NaN</code>.</p></li>
+            <li>
+              <p>
+                If the object's <var>session type</var> is <a def-id="persistent-usage-record-session"></a>, store the <var>record of key usage</var>.
```
ddorwin: minor: Is there a reason you added this step after the two above? It seems like the algorithm should do all the internal work then do the externally-visible steps.

I guess this is consistent with the order in remove(). Perhaps we should reorder both in a separate commit.

ddorwin: In the happy case:
remove() => message => update() => Session Close

That means we will be storing the record for a session we removed.

This step is correct, but we need to avoid executing it in the case where it was triggered by the removal of a session.

ddorwin:
LG overall, but we have an algorithm conflict. I'm not sure how best to resolve that.

Also, you will probably want the final commit message to contain "Fix #..."

encrypted-media-respec.html
mwatson2: ok

encrypted-media-respec.html
mwatson2: On the order of messages / storing data for both remove() and Session Close, this order is not observable, but I agree that it is more readable to have the storage of the *record of key usage* or the *record of license destruction* described next to the destruction of the keys or license respectively. I'll create a separate issue for that.

On the second point, yes, the *record of key usage* or *record of license destruction* should be stored only if keys or licenses were destroyed at that time. Perhaps both the destruction and messaging should be in an "if any license / keys exist..." ? This would also address your first point.

mwatson2:
I addressed both issues here. LMK what you think.

ddorwin:
Thanks. LGTM with nits in comments.

mwatson2:
Merged as afe53abdf8a47bc3e484f5af190af8f4050245a7

Issue #86, opened 2015-09-01
"tracked" session type name is misleading


mwatson2:
There is a danger that the "tracked" session type is mis-understood as being related to user "tracking" of the form recently addressed by the TAG: http://www.w3.org/2001/tag/doc/unsanctioned-tracking/

Since the purpose of this session type is unrelated to unsanctioned user tracking, it should probably have a different name.

Suggestions for bikeshed discussion: "persistent-key-usage", "persistent-usage-record", "persistent-record", "recorded", ...

The original "persistent-release-message" is also an option, but is a bit of a mouthful and doesn't capture what is actually persisted.

ddorwin:
"persistent-usage-record" seems like the most accurate description.

As with the original name, there is the potential for confusion since "persistent-license" sessions might also persist usage records, but the same is true for any of these names. We should try to avoid confusion, but we may need to do that in prose.

mwatson2:
PR: https://github.com/w3c/encrypted-media/pull/90

mwatson2:
If there are no comments on this proposal, I will go ahead and merge it.

ddorwin:
LGTM.

Please see https://lists.w3.org/Archives/Public/public-html-media/2015Sep/0008.html. That will allow you to have a single commit with the correct issue number.

Tip: If you change the message to `Fix #86...` this issue will be [automatically closed](https://help.github.com/articles/closing-issues-via-commit-messages/).

mwatson2:
I believe my git fu has been successfully upgraded and the squashing / rebase accomplished. LMK if anything looks wrong.

mwatson2:
Hmm, whilst there is only one commit in the PR, I see two in the main repository now it is merged - the commit from the PR and the merge. Is that correct ?

ddorwin:
Yes, that is expected when using the green “Merge pull request” button, but we can do better. Here is a better solution: https://lists.w3.org/Archives/Public/public-html-media/2015Sep/0014.html

PR #90, opened 2015-09-03
Issue #86: Rename 'tracked' session type to 'persistent-usage-record'

mwatson2:


mwatson2:
Please not first commit has wrong issue number. Correct issue is #86

PR #94, opened 2015-09-15
Prettify ReSpec config

tripu:
I was about to submit a PR that would add a few changes to enable [automatic publication of the WD with Echidna](https://github.com/w3c/echidna/wiki/FAQ).

One of the necessary changes is to [add the W3C IDs of all editors of the spec](https://github.com/w3c/echidna/wiki/Preparing-your-document#data-editor-id) to the ReSpec config.

Before editing that *very long* line, I thought maybe we could format it better?

This PR tries to do that. I think the change isn't complete; this `script` element has to be moved somewhere else for ReSpec to be able to use it.

Can someone take a look, and confirm this is a good idea? Specifically, that this chunk of JSON isn't automatically overwritten by some tool&hellip;

### DIFF
--- a/index.html
+++ b/index.html
@@ -472,7 +472,548 @@ table.simple {
         display: none;
     }
 }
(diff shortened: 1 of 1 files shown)


mwatson2:
index.html is entirely generated by ReSpec from the source encrypted-media-respec.html. You need to modify the config in that source file, not index.html.

ddorwin:
@tripu Is this really an issue of the ReSpec output? Perhaps it can be updated to avoid such long lines, at least in this area.

I think PR #95 addressed your original goal.

tripu:
Indeed, @ddorwin. Thanks.

PR #95, opened 2015-09-22
Added automatic publishing settings

plehegar:


plehegar:
Regarding other files, they can and should stay in the same repo. We could point to https://github.com/w3c/encrypted-media/commits/gh-pages instead of https://github.com/w3c/encrypted-media/commits/gh-pages/encrypted-media-respec.html if you'd like.

plehegar:
Unless I'm missing something, I think this is good to be merged.

ddorwin:
LGTM. Thanks. I'll manually squash and merge it per our [new process](https://lists.w3.org/Archives/Public/public-html-media/2015Sep/0025.html).

jdsmith3000:
I think index.html exists so that we had modest control over what got viewed on the web (e.g. we could roll up a few changes before updating the index.html copy).  In practice, I think we did it almost every commit.  I think it could be removed.

ddorwin:
The main purpose of having index.html separate from the ReSpec source (encrypted-media-respec.html) is to have the spec at https://w3c.github.io/encrypted-media/ load immediately without the delay and "jarring" of on-the-fly generation. This also ensures that URLs with anchors load to the correct location.

ddorwin:
Merged as 501a9c2cb14bb7b428b98c4281cd5aa2c8fb297d.

jdsmith3000:
Doesn’t this same purpose apply equally to the published version, which we are setting to the ReSpec copy?

ddorwin:
@jdsmith3000, the published version will be static as before. Among other things, the publishing tool needs to use the ReSpec source to generate a WD instead of an ED. I assume that this line from `W3CTRMANIFEST` does that:
`encrypted-media-respec.html?specStatus=WD;shortName=encrypted-media respec`

The result will need to be saved as `index.html` so that http://www.w3.org/TR/encrypted-media/ works. I assume the tool handles that, but @plehegar can confirm.

Issue #108, opened 2015-10-20
Ensure JWK and JWS references are sufficiently up to date


ddorwin:
The uses of JWK were based on more recent versions of those specs that the version of the spec used by ReSpec (https://tools.ietf.org/html/draft-ietf-jose-json-web-key-11). Whatever database ReSpec uses should probably be updated. We should then ensure our uses are compatible.

At the same time, we should get JWS added to that database and remove our `localBiblio` entry for it.

ddorwin:
@plehegar will follow up with either a PR or changes to the database.

ddorwin:
@plehegar has prepared the two PRs above. I will merge them manually per our process and update the commit messages to reference this issue.

PR #114, opened 2015-10-29
JWK is now an RFC

plehegar:
Just making sure to use the latest reference.


ddorwin:
Thanks. LGTM. I will try to land this tomorrow and update the description to reference #108.

ddorwin:
Merged as 05d9fb6.

PR #116, opened 2015-10-29
powerful features is a normative reference

plehegar:
the powerful features spec is missing in the normative reference section despite being used in the spec. This is a proposed fix.


ddorwin:
Merged as 4cfb60414fabf5674efe0ec20aa580f5e116369e.

Issue #48, opened 2015-04-10
Clear Key should not use "A128KW" as the JWK "alg"


ddorwin:
Clear Key's License Format specifies JWK with an "alg" (algorithm) value of "A128KW".

The JSON Web Algorithms (JWA) registry [describes] (https://tools.ietf.org/html/draft-ietf-jose-json-web-algorithms-40#page-39) "A128KW"  as "AES Key Wrap using 128 bit key." However, Clear Key does not use key wrapping, so this is an incorrect description of the algorithm.

According to the JWK specification of the ["alg" parameter](https://tools.ietf.org/html/draft-ietf-jose-json-web-key-41#section-4.4):
> The values used  should either be registered in the
   IANA JSON Web Signature and Encryption Algorithms registry defined in
   [JWA] or be a value that contains a Collision-Resistant Name....
 Use of this member is OPTIONAL.

Rather than attempting to register "A128" or using some ugly Collision-Resistant Name, we should just drop "alg" from the specification. The implementation will use the key as specified by the container anyway.

steelejoe:
This seems reasonable since usage of "alg" is OPTIONAL in the JWK spec as well.

PR #126, opened 2015-11-10
Fix for issue #48.

tinskip:
Removed references to key wrapping algorithm for clearkey.


ddorwin:
Thanks. You need to edit encrypted-media-respec.html then generate index.html. See https://github.com/tinskip/encrypted-media/blob/gh-pages/TEAM.md.

Also, modify the commit message to be of the form:
>Fix #48: <fix>

Where `<fix>` might be `Remove "alg" from Clear Key's License Format.`

PR #136, opened 2015-12-03
Fix #113: Check if session close algorithm has been run in generateRe…

### LINKED ISSUE #113: generateRequest() and load() should check "If the Session Close algorithm has been run"
The `generateRequest()` and `load()` algorithms do not have check "If the Session Close algorithm has been run on this object" like the algorithms for the other `MediaKeySession` methods. This makes sense in normal operation since there are checks of the object's _uninitialized_ value. However, any asynchronous closure (e.g. #102) could occur before these methods are called. Thus, we should add these checks to these algorithms to cover such corner cases.

cpearce:
…quest() and load().

Add text as suggested in issue #113.


ddorwin:
As with the other PR, you'll need to update the ReSpec.

index.html
```diff
@@ -1885,6 +1885,7 @@ <h4 id="h-is-persistent-session-type" resource="#h-is-persistent-session-type"><
             </td></tr><tr><td class="prmName">initData</td><td class="prmType"><code>BufferSource</code></td><td class="prmNullFalse"><span role="img" aria-label="False">✘</span></td><td class="prmOptFalse"><span role="img" aria-label="False">✘</span></td><td class="prmDesc">
               <a href="#initialization-data">Initialization Data</a>
             </td></tr></tbody></table><div><em>Return type: </em><code>Promise&lt;void&gt;</code></div><p>When this method is invoked, the user agent must run the following steps:</p><ol class="method-algorithm">
+            <li><p>If the <a href="#session-close">Session Close</a> algorithm has been run on this object, return a promise rejected with a new <code><a href="https://heycam.github.io/webidl/#dfn-DOMException">DOMException</a></code> whose name is <code><a href="#dfn-InvalidStateError">InvalidStateError</a></code>.</p></li>
```
ddorwin: In the existing cases, this step is after checking _callable_. We should try to be consistent by either running this step after checking _uninitialized_ OR making this the first step in all cases.

I'm leaning slightly towards the latter. What do you think?

cpearce:
Making it the first step in call cases makes sense. I'll update the PR (again),

ddorwin:
I didn't get an email when you committed, so I didn't see that you had made the change. If you add a comment after uploading a new commit, we'll know to look.

Mark fixed #113 today, but I merged your PR - modified to only be for the ordering consistency - as 6d3382e01e16bed9d877acc25e52c21ab722404a

Issue #100, opened 2015-10-19
Is "running the Encrypted Block Encountered algorithm" the correct way to Attempt to Resume Playback If Necessary?


ddorwin:
The Initialization Data Encountered and Encrypted Block Encountered algorithms have steps to "Wait for a signal to resume playback."

This wait prevents further processing of the media data, which is an easy way to ensure we know the state of playback. However, this entirely blocks the resource fetch algorithm, and it's weird to block in an algorithm. So, the first question is whether this is correct.

Second, the Attempt to Resume Playback If Necessary algorithm "Attempt[s] to resume the resource fetch algorithm by running the Encrypted Block Encountered algorithm." Running this algorithm while currently waiting in the algorithm doesn't seem correct. It seems even more incorrect when waiting in a different algorithm, which is what happens when the `setMediaKeys()` algorithm invokes Attempt to Resume Playback If Necessary algorithm after the Initialization Data Encountered algorithm ends in the waiting step. It's also not clear that running the Encrypted Block Encountered algorithm is the next logical step after unblocking the Initialization Data Encountered algorithm. (Note that it is also possible that `setMediaKeys()` would unblock the wait in the Encrypted Block Encountered algorithm in some implementations.)

We should a) consider another way of stalling the processing of media data and b) determine the best way to resume based on the outcome of (a).

mwatson2:
I think the solution is to suspend the resource fetch algorithm instead of blocking in our Encrypted Block Encountered algorithm. Then the "attempt to resume playback" would check whether it (the resource fetch algorithm) can resume (by explicitly checking the state of the key that is needed) and move it back out of the suspended state.

The resource fetch algorithm already has a "suspended" state:

*User agents may decide to not download more content at any time, e.g. after buffering five minutes of a one hour media resource, while waiting for the user to decide whether to play the resource or not, while waiting for user input in an interactive resource, or when the user navigates away from the page. When a media element's download has been suspended, the user agent must queue a task, using the DOM manipulation task source, to set the networkState to NETWORK_IDLE and fire a simple event named suspend at the element. If and when downloading of the resource resumes, the user agent must queue a task to set the networkState to NETWORK_LOADING. Between the queuing of these tasks, the load is suspended (so progress events don't fire, as described above).*

[...]

Issue #129, opened 2015-11-25
Confirm `readyState` behavior when blocked waiting for a key is correct


ddorwin:
The value of `readyState` when blocked waiting for a key (causing the `waitingforkey` event) was raised by @cpearce-mozilla in #100. Although the EME spec is explicit about the expected behavior, it is not clear whether that is the correct behavior. This issue tracks confirming the specified behavior or modifying it to be correct. This likely depends on correctly interpreting the definition of `readyState`.

The EME spec currently says:
>The element is potentially playing and its `readyState` is equal to `HAVE_FUTURE_DATA` or greater.

This was based on an interpretation of `readyState` as relating only to the buffering of media data (see #7), which seems to be implied by the definitions of the [Ready states](https://html.spec.whatwg.org/multipage/embedded-content.html/#ready-states).

Among the questions to address are:
* Should being blocked on a key affect the media element's `readyState`?
  * Note that this would cause the HTMLMediaElement `waiting` and `canplay` events to be fired in response to key availability (see #7). In #7, we decided to avoid using these events and only fire an EME-specific event.
* Should the `readyState` indeed be "equal to `HAVE_FUTURE_DATA` or greater" when the Queue a "waitingforkey" Event algorithm is run?
  * More specifically, is the `readyState` "equal to `HAVE_FUTURE_DATA` or greater" when the Encrypted Block Encountered algorithm is run?

<br/>The HTML spec [says](https://html.spec.whatwg.org/multipage/embedded-content.html#ready-state):
>Media elements have a _ready state_, which describes to what degree they are ready to be rendered at the current playback position.

Based on that definition ("ready to be rendered"), it sounds like `readyState` should consider whether the media data can currently be rendered, which would include decryption _and decoding_. Furthermore, `HAVE_CURRENT_DATA` and higher would imply that the current data has already been (decrypted and) decoded. However, I do not believe that is the intent, so perhaps the definition needs to be corrected.

The [`HAVE_ENOUGH_DATA` definition](https://html.spec.whatwg.org/multipage/embedded-content.html#dom-media-have_enough_data) actually references the fetching of the data, which would seem to imply that `readyState` is indeed related to buffering (and the resource fetch algorithm).

If the "ready to be rendered" definition is correct, the definition of `HAVE_ENOUGH_DATA` should probably also include decoding:
> The user agent estimates that data is being fetched **and decoded** at a rate...

[...]

PR #139, opened 2015-12-15
Issue #100: PR to revise the way we monkey-patch HTML media algorithms

mwatson2:


encrypted-media-respec.html
```diff
@@ -2024,9 +2024,21 @@
...
       <h2><a>HTMLMediaElement</a> Extensions</h2>
+      
+      <!-- Put all the monkey-patching here -->
+      
       <p>This section specifies additions to and modifications of the <a def-id="htmlmediaelement"></a> [[!HTML5]] when the Encrypted Media Extensions are supported.</p>
+      
+      <p>This specification adds two internal values to the <a def-id="htmlmediaelement"></a>: <var>waiting for key</var> which shall have a boolean value and <var>encrypted block queue</var> which shall be a queue of encrypted blocks awaiting decryption.</p>
 
-      <p>When a <a def-id="htmlmediaelement"></a> is created, its internal <var>waiting for key</var> value is initialized to <code>false</code>.</p>
+      <p>When a <a def-id="htmlmediaelement"></a> is created, its internal <var>waiting for key</var> value shall initialized to <code>false</code> and its <var>encrypted block queue</var> value shall be empty.</p>
```
cpearce: "<var>waiting for key</var> value shall *be* initialized"

i.e. insert "be" between "shall" and "initialized".

encrypted-media-respec.html
```diff
@@ -2024,9 +2024,21 @@
 
     <section>
       <h2><a>HTMLMediaElement</a> Extensions</h2>
+      
+      <!-- Put all the monkey-patching here -->
+      
       <p>This section specifies additions to and modifications of the <a def-id="htmlmediaelement"></a> [[!HTML5]] when the Encrypted Media Extensions are supported.</p>
+      
+      <p>This specification adds two internal values to the <a def-id="htmlmediaelement"></a>: <var>waiting for key</var> which shall have a boolean value and <var>encrypted block queue</var> which shall be a queue of encrypted blocks awaiting decryption.</p>
```
ddorwin: nits:
* It might be nice to use a list.
* Add commas around "which" fragments.

encrypted-media-respec.html
```diff
@@ -2024,9 +2024,21 @@
...
 
-      <p>When a <a def-id="htmlmediaelement"></a> is created, its internal <var>waiting for key</var> value is initialized to <code>false</code>.</p>
+      <p>When a <a def-id="htmlmediaelement"></a> is created, its internal <var>waiting for key</var> value shall initialized to <code>false</code> and its <var>encrypted block queue</var> value shall be empty.</p>
+      
+      <p>In addition to the criteria specified in [[!HTML5]] an <a def-id="htmlmediaelement"></a> shall be considered a <a def-id="blocked-media-element"></a> if its <var>waiting for key</var> value is <code>true</code>.</p>
```
ddorwin: nit: comma before "an"

encrypted-media-respec.html
```diff
@@ -2024,9 +2024,21 @@
...
+      
+      <p>When the media element encounters <a def-id="initialization-data"></a> in the <a def-id="media-data"></a> during the <a def-id="resource-fetch-algorithm"></a> the user agent shall run the <a def-id="initdata-encountered-algorithm"></a> algorithm.</p>
```
ddorwin: nits:
* Add a comma before "the user agent"
* The media element encounters the initData, but the user agent performs the actions. This seems inconsistent. The former comes from the existing text below, but that may also have been incorrect.

encrypted-media-respec.html
```diff
@@ -2024,9 +2024,21 @@
...
+      
+      <p>For each block of encrypted <a def-id="media-data"></a> encountered during the <a def-id="resource-fetch-algorithm"></a> the user agent shall run the <a def-id="encrypted-block-encountered-algorithm"></a> algorithm in the order the encrypted blocks were encountered.</p>
```
ddorwin: nit: Add a comma before "the user agent."

encrypted-media-respec.html
```diff
@@ -2024,9 +2024,21 @@
...
+      <p class="note">The above requirement provides flexibility for user agent implementations to perform decryption at any time after an encrypted block is encountered before it is needed for playback.</p>
```
ddorwin: s/requirement/step/?

encrypted-media-respec.html
```diff
@@ -2216,9 +2228,21 @@
 
         <section id="encrypted-block-encountered">
           <h4>Encrypted Block Encountered</h4>
-          <p>The following steps are run when the media element encounters a block of encrypted <a def-id="media-data"></a> during the <a def-id="resource-fetch-algorithm"></a>:</p>
-
+          <p>This algorithm queues a block of encrypted media data, <var>block</var> for decryption and attempts to decrypt if possible. The following steps are run:</p>
```
ddorwin: nit: comma after `<var>block</var>`.
Or define the variable in a step: `Let <var>block</var> be the block of encrypted media data.`

encrypted-media-respec.html
```diff
@@ -2303,10 +2334,13 @@
             </li>
             <li>
               <p>Run the following steps:</p>
-              <p class="note">These steps are reached when there is no usable key for the block.</p>
+              <p class="note">These steps are reached when there is no usable key for <var>block</var>.</p>
               <ol>
                 <li><p>Run the <a def-id="queue-waitingforkey-algorithm"></a> algorithm on the media element.</p></li>
-                <li><p>Wait for a signal to resume playback.</p></li>
+                <!-- I propose removing the following step, as it is odd to "wait" inside an algorith. The media element is now a "blocked media element" and will become (not one of those)
```
ddorwin: Sounds fine. We might add a Note explaining what happens.

encrypted-media-respec.html
```diff
@@ -2331,15 +2365,16 @@
...
             It should only be called when the <a def-id="htmlmediaelement"></a> object is <a def-id="potentially-playing"></a> and its <a def-id="readystate"></a> is equal to <a def-id="have-future-data"></a> or greater.  
             Requests to run this algorithm include a target <a def-id="htmlmediaelement"></a> object.
           </p>
           <p>The following steps are run:</p>
           <ol>
             <li><p>Let the <var>media element</var> be the specified <a def-id="htmlmediaelement"></a> object.</p></li>
-            <li><p>If the <var>media element</var>'s <var>waiting for key</var> value is <code>false</code>, <a def-id="queue-a-task-to-fire-an-event-named"></a> <a def-id="waitingforkey"></a> at the <var>media element</var>.</p></li>
+            <li><p>If the <var>media element</var>'s <var>waiting for key</var> value is <code>true</code>, abort these steps.
             <li><p>Set the <var>media element</var>'s <var>waiting for key</var> value to <code>true</code>.</p></li>
+            <li><p><a def-id="queue-a-task-to-fire-an-event-named"></a> <a def-id="waitingforkey"></a> at the <var>media element</var>.</p></li>
```
ddorwin: Capital 'Q' - there are two macros.

encrypted-media-respec.html
```diff
@@ -2355,8 +2390,13 @@
           <ol>
             <li><p>Let the <var>media element</var> be the specified <a def-id="htmlmediaelement"></a> object.</p></li>
             <li><p>If the <var>media element</var>'s <var>waiting for key</var> is <code>false</code>, abort these steps.</p></li>
-            <li><p>Attempt to resume the <a def-id="resource-fetch-algorithm"></a> by running the <a def-id="encrypted-block-encountered-algorithm"></a> algorithm.</p></li>
-            <li><p>If the user agent can advance the <a def-id="current-playback-position"></a> in the <a def-id="direction-of-playback"></a>, set the <var>media element</var>'s <var>waiting for key</var> value to <code>false</code>.</p></li>
+            <li><p>Run the <a def-id="attempt-to-decrypt-algorithm"></a> algorithm on the <var>media element</var>.</p></li>
+            <li><p>If the user agent can advance the <a def-id="current-playback-position"></a> in the <a def-id="direction-of-playback"></a>:</p>
+              <ol>
+                <li><p>Set the <var>media element</var>'s <var>waiting for key</var> value to <code>false</code>.</p></li>
+                <li><p>Resume playback</p></li>
```
ddorwin: This step seems less precise than the rest of the patching. I wonder if there is more appropriate text.

nit: Missing period.

index.html
```diff
@@ -2889,22 +2923,23 @@ <h4 id="h-encrypted-block-encountered" resource="#h-encrypted-block-encountered"
...
           </p>
           <p>The following steps are run:</p>
           <ol>
             <li><p>Let the <var>media element</var> be the specified <code><a href="http://www.w3.org/TR/html5/embedded-content-0.html#htmlmediaelement">HTMLMediaElement</a></code> object.</p></li>
-            <li><p>If the <var>media element</var>'s <var>waiting for key</var> value is <code>false</code>, <a href="http://www.w3.org/TR/html5/webappapis.html#queue-a-task">queue a task</a> to <a href="http://www.w3.org/TR/html5/webappapis.html#fire-a-simple-event">fire a simple event</a> named <code><a href="#dom-evt-waitingforkey">waitingforkey</a></code> at the <var>media element</var>.</p></li>
-            <li><p>Set the <var>media element</var>'s <var>waiting for key</var> value to <code>true</code>.</p></li>
+            <li><p>If the <var>media element</var>'s <var>waiting for key</var> value is <code>true</code>, abort these steps.
+            </p></li><li><p>Set the <var>media element</var>'s <var>waiting for key</var> value to <code>true</code>.</p></li>
```
ddorwin: Line break is in the wrong place.

encrypted-media-respec.html
```diff
@@ -2331,15 +2365,16 @@
 
         <section id="queue-waitingforkey">
           <h4>Queue a "waitingforkey" Event</h4>
-          <p>The Queue a "waitingforkey" Event algorithm is run when the CDM cannot decrypt the current block.
+          <p>The Queue a "waitingforkey" Event algorithm is run when the CDM cannot decrypt an encrypted block.
             It should only be called when the <a def-id="htmlmediaelement"></a> object is <a def-id="potentially-playing"></a> and its <a def-id="readystate"></a> is equal to <a def-id="have-future-data"></a> or greater.
```
ddorwin: I'm not sure this statement is correct anymore. I think we allow it to be called when `<var>waiting for key</var>` is true (there is an early exit below), which would mean it is blocked, which means it is not potentially playing.

If we want to ensure `<var>waiting for key</var>` is not true when this is called (makes sense given the name of the algorithm) then this statement would be correct.

encrypted-media-respec.html
```diff
@@ -2331,15 +2365,16 @@
...
+          <p>The Queue a "waitingforkey" Event algorithm is run when the CDM cannot decrypt an encrypted block.
             It should only be called when the <a def-id="htmlmediaelement"></a> object is <a def-id="potentially-playing"></a> and its <a def-id="readystate"></a> is equal to <a def-id="have-future-data"></a> or greater.  
             Requests to run this algorithm include a target <a def-id="htmlmediaelement"></a> object.
           </p>
           <p>The following steps are run:</p>
           <ol>
             <li><p>Let the <var>media element</var> be the specified <a def-id="htmlmediaelement"></a> object.</p></li>
-            <li><p>If the <var>media element</var>'s <var>waiting for key</var> value is <code>false</code>, <a def-id="queue-a-task-to-fire-an-event-named"></a> <a def-id="waitingforkey"></a> at the <var>media element</var>.</p></li>
+            <li><p>If the <var>media element</var>'s <var>waiting for key</var> value is <code>true</code>, abort these steps.
             <li><p>Set the <var>media element</var>'s <var>waiting for key</var> value to <code>true</code>.</p></li>
```
ddorwin: Should we add a Note that the media element is now a "blocked media element" and thus not "potentially playing"?
(I suggested this above where this algorithm is called. It is the setting of `<var>waiting for key</var>` to true that triggers this, though. Perhaps we should move all these steps to the call point. This would also address my comment immediately above.

The other use of this algorithms is in the Initialization Data Encountered algorithm. I believe the state and Note are appropriate for that path as well, so maybe we should keep/put them here. We may want to rename this algorithm, though.

encrypted-media-respec.html
```diff
@@ -2355,8 +2390,13 @@
           <ol>
             <li><p>Let the <var>media element</var> be the specified <a def-id="htmlmediaelement"></a> object.</p></li>
             <li><p>If the <var>media element</var>'s <var>waiting for key</var> is <code>false</code>, abort these steps.</p></li>
-            <li><p>Attempt to resume the <a def-id="resource-fetch-algorithm"></a> by running the <a def-id="encrypted-block-encountered-algorithm"></a> algorithm.</p></li>
-            <li><p>If the user agent can advance the <a def-id="current-playback-position"></a> in the <a def-id="direction-of-playback"></a>, set the <var>media element</var>'s <var>waiting for key</var> value to <code>false</code>.</p></li>
+            <li><p>Run the <a def-id="attempt-to-decrypt-algorithm"></a> algorithm on the <var>media element</var>.</p></li>
```
ddorwin: The intro paragraph to this algorithm is not accurate. It can also be called from setMediaKeys() to unblock the wait in the Initialization Data Encountered algorithm.

encrypted-media-respec.html
ddorwin: There are probably some crazy paths if an implementation allowed clearing/changing the MediaKeys during playback. Perhaps we should reset these internal values in the setMediaKeys() algorithm. (Note that it also sets them if necessary.)

cpearce: Potentially script could create a new MediaKeys for the key for which playback is blocked waiting on. Seems unlikely that would be desirable though. But if that was desired, we'd want to not be clearing the queue of encrypted blocks awaiting decryption in the setMediaKeys algorithm, so the new MediaKeys could decode where the old one left off.

Neither Firefox nor Chrome (last time I checked) supports changing the MediaKeys after playback has started. It would be a bit of a hassle for us to implement, so we're unlikely to bother to implement it unless a compelling use case comes up.

encrypted-media-respec.html
cpearce: Firefox won't advance past HAVE_CURRENT_DATA until we have successfully decoded the data at the current playback position, i.e. until we've had usable keys and used them. So we wouldn't be able to fire this event... uh... ever, because not having the key for the current playback position means our readyState we'll drop back below HAVE_FUTURE_DATA.

Can we remove the HAVE_FUTURE_DATA check? What is the readyState check trying to achieve that the potentially playing check and the checks before calling this algorithm don't achieve?

mwatson2:
@ddorwin @cpearce Thanks for the detailed review comments. I had hoped to get to implementation this week, but now it looks like it may be next week.

encrypted-media-respec.html
mwatson2: Given the implementation status, would it be better to remove altogether the ability to change (and clear ?) MediaKeys during playback ? I think the only use-case would be seamless playback across changes in keysystem or codec / robustness combinations which seem like things that are very unlikely to be widely supported any time soon.

mwatson2: On reviewing the text, I don't see a need to update the state in setMediaKeys. That method runs the "Attempt to resume playback if necessary" method, so if we are blocked waiting for a key at this time then this will unblock us (if the right key is there).

encrypted-media-respec.html
mwatson2: I don't think we need the explicit "resume playback" step. I will add notes to the places that the *waiting for key* variable is modified to describe how these changes change the "blocked media element" state an the consequent effects on playback.

encrypted-media-respec.html
ddorwin: I believe the idea was that it doesn't make sense to be waiting for a key if you don't have the data necessary to advance. The ready states are strictly about buffering. This was discussed in #100 before this specific question was migrated to #129.

mwatson2: On Tue, Feb 2, 2016 at 3:42 PM, ddorwin <notifications@github.com> wrote:

> In encrypted-media-respec.html
> <https://github.com/w3c/encrypted-media/pull/139#discussion_r51656285>:
>
> > @@ -2331,15 +2365,16 @@
> >
> >          <section id="queue-waitingforkey">
> >            <h4>Queue a "waitingforkey" Event</h4>
> > -          <p>The Queue a "waitingforkey" Event algorithm is run when the CDM cannot decrypt the current block.
> > +          <p>The Queue a "waitingforkey" Event algorithm is run when the CDM cannot decrypt an encrypted block.
> >              It should only be called when the <a def-id="htmlmediaelement"></a> object is <a def-id="potentially-playing"></a> and its <a def-id="readystate"></a> is equal to <a def-id="have-future-data"></a> or greater.
>
> I believe the idea was that it doesn't make sense to be waiting for a key
> if you don't have the data necessary to advance. The ready states are
> strictly about buffering. This was discussed in #100
> <https://github.com/w3c/encrypted-media/issues/100> before this specific
> question was migrated to #129
> <https://github.com/w3c/encrypted-media/issues/129>.
>
​Do you mean to say that "... cannot decrypt the current block." is the
correct language ? I thought I had merged that change correctly from the
main branch. Let me check.​

encrypted-media-respec.html
```diff
@@ -2028,9 +2028,29 @@
 
     <section>
       <h2><a>HTMLMediaElement</a> Extensions</h2>
-      <p>This section specifies additions to and modifications of the <a def-id="htmlmediaelement"></a> [[!HTML5]] when the Encrypted Media Extensions are supported.</p>
-
-      <p>When a <a def-id="htmlmediaelement"></a> is created, its internal <var>waiting for key</var> value is initialized to <code>false</code>.</p>
+      
+      <!-- Put all the monkey-patching here -->
+      
+      <p>This section specifies additions to and modifications of the <a def-id="htmlmediaelement"></a> [[!HTML5]] when the Encrypted Media Extensions are supported:</p>
```
ddorwin: nit: This is a statement so it should end in a period. Since the list does not contain all modifications/additions (there are subsections), it doesn't make sense to reword to keep the colon.

When I suggested a list, I was referring to a list of the new variables. One option to address the nit and increase readability would be to have two separate lists - one of internal values and one of modifications to existing values and algorithms. I'm open to other options too.

encrypted-media-respec.html
ddorwin: @mwatson2, I was replying to @cpearce's comment, not the actual text (even though that is apparently how GitHub shows the comment).

Thinkin about it more, it's not clear that `HAVE_CURRENT_DATA` means the data has actually been decoded or displayed. If not, perhaps `HAVE_CURRENT_DATA` is the minimum state.

ddorwin:
LGTM with my comments above. When you squash to commit, please update the final commit message to "**Fix** #100..."

mwatson2:
Merged as a9d74c124bf12402cb5dbe91485d2340d38b6da1

Issue #140, opened 2015-12-17
Typo in HTMLMediaElement.setMediaKeys method definition


cpearce:
The setMediaKeys method definition, step 5.1 reads:

"If mediaKeys is not null, it is already in use by another media element, and the user agent is unable to use it with this element, let this object's attaching media keys value be false and reject promise with a new DOMException whose name is QuotaExceededError."

I think that's a typo, it should read:

"If mediaKeys is not null, **and** it is already in use by another media element, and the user agent is unable to use it with this element, let this object's attaching media keys value be false and reject promise with a new DOMException whose name is QuotaExceededError."

ddorwin:
Those are three separate conditions that must all be true. It follows the structure `A, B, and C.` In this case, B implies A, but I think it is still helpful to be explicit.

I think "it" could be less ambiguous, and that might help. "It" might also be technically incorrect.

I propose the following, which is consistent with step 5.2.3.
>If mediaKeys is not null, **the CDM instance represented by _mediaKeys_** is already in use...

cpearce:
OK, that sounds fine. I've done that in PR#141.

PR #141, opened 2015-12-17
Issue #140: Reduce potential for ambiguity in setMediaKeys definition.

cpearce:
One way to interpret the current setMediaKeys method definition is
if the method is passed a non-null MediaKeys that it should assume
the MediaKeys is in use by another HTMLMediaElement. We can tweak
the wording to make it clear that's not the case.


encrypted-media-respec.html
```diff
@@ -2067,7 +2067,7 @@
             <li><p>Let <var>promise</var> be a new promise.</p></li>
             <li><p>Run the following steps in parallel:</p>
               <ol>
-                <li><p>If <var>mediaKeys</var> is not null, it is already in use by another media element, and the user agent is unable to use it with this element, let this object's <var>attaching media keys</var> value be false and reject <var>promise</var> with <a def-id="new-domexception-named"></a> <a def-id="QuotaExceededError"></a>.</p></li>
+                <li><p>If <var>mediaKeys</var> is not null, the CDM instance represented by mediaKeys it is already in use by another media element, and the user agent is unable to use it with this element, let this object's <var>attaching media keys</var> value be false and reject <var>promise</var> with <a def-id="new-domexception-named"></a> <a def-id="QuotaExceededError"></a>.</p></li>
```
ddorwin: `<var>mediaKeys</var>`

ddorwin: Remove "it ".

ddorwin:
Thanks. Two minor comments then LG. Please also update the description to "Fix #..." so the issue will automatically be closed when merged.

cpearce:
Oops, I've pushed a commit to address those comments.

ddorwin:
Merged as d2cccea.

Issue #2, opened 2014-12-12
stack_overview.svg needs to be updated to match the latest spec


ddorwin:
* needkey => encrypted
* isTypeSupported() => requestMediaKeySystemAccess(), createMediaKeys(), createSession()
* createSession() => generateRequest()

We should also add the new objects next to MediaKeySession.

steelejoe:
I am working on this now.

steelejoe:
The section labeled "Optional" does not appear to be optional -- I would expect an "encrypted" event to always be sent for an encrypted media. Is this not correct?

steelejoe:
This is the updated SVG. Unfortunately generating a pull request failed for me -- have to dig more into that. [stack_overview.svg.zip](https://github.com/w3c/encrypted-media/files/86824/stack_overview.svg.zip)

ddorwin:
Thanks, @steelejoe.

You are correct that firing "encrypted" is not optional. What is optional is whether the application handles it or provides data such that it fires before checking for supported types. Previously, checking the supported types was also optional, though advisable, previously. (The original spec said, "Likely only one of the arrows in the dotted box will occur.")

Now, requestMediaKeySystemAccess() is certainly required. The only thing that's optional is whether to handle "encrypted" and whether it occurs before or after any of the EME method calls. The dashed optional arrow is probably sufficient.

Nits:
* For requestMedia...KeySystemAccess(), use a hyphen like in a newspaper.
* The MediaKeySession boxes changed colors? Is this intentional? They should be distinct from other boxes.
* The far left arrow to Net should be a little shorter (it was wrong in the current version).
* The Generate Request arrows should be to the right of the generateRequest() arrows.
* There is something strange about the head of the second update() arrow. It looks like there might be two arrows there, one of which is offset.
* The first Generate Request arrow no longer touches the CDM box.

MediaKeySystemAccess and MediaKeys are not shown (and the arrows point to the MediaStack box) while MediaKeySession is. Adding those objects could complicate the image, so we may just want to leave it as is. One option might be to add them above the MediaKeySession boxes. MediaKeys would need to stretch to the other createSession() anyway.

paulbrucecotton:
Can one of the EME editors (@ddorwin, @mwatson2 or @jdsmith3000) please review and commit Joe's changes?

mwatson2:
[...]

PR #143, opened 2016-01-14
Issue #2 - Updated diagram with latest methods

steelejoe:
Updating stack-overview diagram to reflect latest methods


ddorwin:
Thanks. LG except for some visual issues and nits. Please also change the commit message to be "Fix #2: Update the figure" or something like that. The important part is "Fix #2:"

Unless otherwise noted, these issues appear in both Chrome and Firefox.
* In Firefox, text is no longer centered. Examples: CDN, Net, MediaKeySession, CDM. Also, Media Stack runs into the first MediaKeySession.
* The first MediaKeySession is not centered even in Chrome. In Firefox, it extends outside its box.
* The font changed. The main problem is that the italics text looks bad, especially the e's.
* The MediaKeySession boxes appear to be at an angle. When scaled up, the horizontal lines are not straight.
* The optional (right) update arrow still has a weird double head.
* Please make the heads of the left update() arrow and the top (up) Frame arrow touch their target objects.

ddorwin:
* The MediaKeySession boxes are different sizes. I think the right one can be trimmed on the right.

steelejoe:
I think this has everything you wanted. You are correct the fonts changed, because the original SVG (as far as I can tell) used outlining which look snice but does not preserver font information. So I made a best guess. I believe the wonky arrows heads should all be corrected as well.

ddorwin:
LGTM. Thanks!

Is there a reason the title was changed back? Perhaps when you force pushed the new commit, replacing the previous one?

There are a couple minor issues, but we can fix those in a separate PR if you want:
* The MediaKeySession boxes appear to be at an angle. When scaled up, the horizontal lines are not straight. This is more pronounced in Chrome that in Firefox.
* The `Frame` up arrow appears to have a double head.

steelejoe:
The title change was likely the force commit. I am not seeing the slant you are seeing on Chrome or Firefox, but I will test on other platforms. Let me put together another PR.

ddorwin:
I can't perceive see a slant, but the horizontal lines are not solid as if they are at an angle. Like `~~~----___` but more subtle. This is more obvious at high resolutions.

ddorwin:
@steelejoe, what do you want to do? Shall we commit this and continue to fix minor details later?

ddorwin:
@steelejoe, should we land this and make improvements later?

steelejoe:
Sorry was looking into the wrong place for updates.  Yes please. Let's
commit this and I will work on correcting the slant after.

ddorwin:
Merged as 336e7bd. Thank you!

PR #147, opened 2016-02-16
Plh/initdata registry

plehegar:
Basically, all the files would appear under https://www.w3.org/html/media/

### DIFF
--- a/cenc-format.html
+++ b/cenc-format.html
@@ -1,6 +1,6 @@
 <!DOCTYPE html>
 <html lang="en" dir="ltr" typeof="bibo:Document " prefix="bibo: http://purl.org/ontology/bibo/ w3p: http://www.w3.org/2001/02pd/rec54#">
-<head><meta lang="" property="dc:language" content="en">
+<head><meta property="dc:language" content="en" lang="">
     <meta http-equiv="Content-Type" content="text/html; charset=utf-8">
     <title>ISO Common Encryption EME Stream Format and Initialization Data</title>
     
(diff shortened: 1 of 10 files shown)


cenc-format.html
```diff
@@ -382,157 +374,92 @@ <h2 id="w3c-editor-s-draft-10-december-2015"><abbr title="World Wide Web Consort
...
       It also defines a common SystemID and PSSH box format for use with Encrypted Media Extensions.
       </p>
-    </section><section id="sotd" class="introductory"><h2 id="h-sotd" resource="#h-sotd"><span property="xhv:role" resource="xhv:heading">Status of This Document</span></h2>
+    </section><section id="sotd" class="introductory"><h2 resource="#h-sotd" id="h-sotd"><span property="xhv:role" resource="xhv:heading">Status of This Document</span></h2>
   
     
       
         <p>
-          <em>This section describes the status of this document at the time of its publication. Other documents may supersede this document. A list of current <abbr title="World Wide Web Consortium">W3C</abbr> publications and the latest revision of this technical report can be found in the <a href="http://www.w3.org/TR/"><abbr title="World Wide Web Consortium">W3C</abbr> technical reports index</a> at http://www.w3.org/TR/.</em>
+          This document is merely a <abbr title="World Wide Web Consortium">W3C</abbr>-internal  document. It
```
ddorwin: The use of specStatus: "base" in ReSpec causes the SotD to say this is an internal document with no standing. At least the former seems incorrect.

cenc-format.html
```diff
@@ -382,157 +374,92 @@ <h2 id="w3c-editor-s-draft-10-december-2015"><abbr title="World Wide Web Consort
...
-          
-          
-          
-          <p>
-            
-              This document was produced by
-              
-              a group
-               operating under the
-              <a id="sotd_patent" property="w3p:patentRules" href="http://www.w3.org/Consortium/Patent-Policy-20040205/">5 February 2004 <abbr title="World Wide Web Consortium">W3C</abbr> Patent
```
ddorwin: The use of specStatus: "base" in ReSpec means there is no patent policy for the document. Is that correct/desirable?

cenc-format.html
```diff
@@ -302,23 +301,16 @@
...
-      <dt>This version:</dt>
-      <dd><a class="u-url" href="https://w3c.github.io/encrypted-media/cenc-format.html">https://w3c.github.io/encrypted-media/cenc-format.html</a></dd>
-      <dt>Latest published version:</dt>
-      <dd><a href="http://www.w3.org/TR/encrypted-media/">http://www.w3.org/TR/encrypted-media/</a></dd>
-    
     
       <dt>Latest editor's draft:</dt>
-      <dd><a href="https://w3c.github.io/encrypted-media/cenc-format.html">https://w3c.github.io/encrypted-media/cenc-format.html</a></dd>
-    
+      <dd><a href="https://www.w3.org/html/media/cenc-format.html">https://www.w3.org/html/media/cenc-format.html</a></dd>
```
ddorwin: The use of specStatus: "base" in ReSpec causes there to be no latest-but-not-yet-published URL.
a) Will these be auto-published?
b) Will they always be auto-published even after EME moves to CR such that there is no need for a "latest" URL?

cenc-format.html
plehegar: Good catch. I need to change that.

PR #161, opened 2016-04-13
fixed a typo

Sean-X:


ddorwin:
Thank you. `index.html` is generated - the source is `encrypted-media-respec.html`. You can either fix and upload that file or I can make the change directly.

Sean-X:
thanks for your comment, I've updated it.

ddorwin:
Merged as c8292ad359f53d1ca495df52c2a0a89b39cb77c5.

Issue #151, opened 2016-03-07
Clarify the meaning of "do[es] not support" a media capability


ddorwin:
Step 12 of the [_Get Supported Capabilities for Audio/Video Type_](https://w3c.github.io/encrypted-media/#get-supported-capabilities-for-audio-video-type) algorithm says
>If the user agent and implementation do not support playback of encrypted media data for the combination of _container_, _media types_, _robustness_ and _local accumulated configuration_, continue to the next iteration.

In practice, _media types_ means codecs. The question is whether "do not support" means (a) "is **certain it does not** support" or (b) "is **not certain** it can support." This is mainly relevant when a codec string is ambiguous or otherwise not known to be supported. These are likely the cases where `canPlayType()` would return `"maybe"` rather than `""` or `"probably"`.

If (a), ambiguous or potentially-supported codec strings would cause the current configuration to be selected. If (b), they would not be included and may cause the current configuration to be skipped.

While (a) might be seen as friendly, it also:
* Prevents an application from using certain patterns.
  * For example, one configuration may include a codec string that indicates advanced features (about which some UAs may not be certain) may be followed by a configuration containing a more common codec string.
  * If the UA does not reject the first codec string, the application cannot differentiate.
* Encourages bad practices (i.e. using incomplete codec strings), which could prevent implementations from providing accurate and specific responses.
  * By being strict in a new API, we can avoid some of the issues we are now stuck with for `canPlayType()`.

We have been more "strict" than friendly before (i.e. requiring all `sessionTypes` to be supported) in order to enable application control similar to the example above.

Note: Part of the issue here is the liberal handling of codec strings (e.g. by `canPlayType()`) in many implementations, that the differentiation between `"maybe"` and `"probably"` is often irrelevant, and that the EME API has two states instead of three. MSE’s `isTypeSupported()`  also only has two states and appears to be liberal, but it does not explicitly need to support a series of configurations like EME does. If it was possible, being strict would probably be better there as well. Perhaps more importantly, `canPlayType()` and `isTypeSupported()` are designed to allow the application to walk a tree of capabilities whereas EME is designed to check lists of specific codecs and get a response. Furthermore, EME allows a subset of the listed types to be selected (and thus satisfy the configuration).

The following are some examples that frequently occur today for `canPlayType()` and MSE's `isTypeSupported()`:
* `"avc1"` - [valid RFC6381 syntax](https://tools.ietf.org/html/rfc6381#section-3.4) but very ambiguous.
[...]

PR #163, opened 2016-04-14
PR for Issue #151: require definite support of media type in requestMediaKeySystemAccess

mwatson2:
…ediaKeySystemAccess

### DIFF
--- a/encrypted-media-respec.html
+++ b/encrypted-media-respec.html
@@ -839,37 +839,42 @@
                     <p class="note">For example, if <var>audio/video type</var> is Video and the top-level type is not "video" or <var>media types</var> contains non-video codecs.</p>
                   </li>
                   <li><p>If <var>robustness</var> is not the empty string and contains an unrecognized value or a value not supported by <var>implementation</var>, continue to the next iteration. String comparison is case-sensitive.</p></li>
-                  <li><p>If the user agent and <var>implementation</var> do not support playback of encrypted <a def-id="media-data"></a> for the combination of <var>container</var>, <var>media types</var>, <var>robustness</var> and <var>local accumulated configuration</var>, continue to the next iteration.</p>
+                  <li><p>If the user agent and <var>implementation</var> definately support playback of encrypted <a def-id="media-data"></a> for the combination of <var>container</var>, <var>media types</var>, <var>robustness</var> and <var>local accumulated configuration</var>:</p>
                     <p class="note"><var>requested media capability</var> (content type and robustness) must be supported with all previously added requested media capabilities.</p>
-                  </li>
-                  <li>
-                    <p>
-                      Add <var>requested media capability</var> to <var>supported media capabilities</var>.
-                    </p>
-                    <p class="note">
-                      This step ensures that the values of the members of entries in <var>supported media capabilities</var> are exactly the strings supplied in
-                      <var>requested media capability</var> without modification by the User Agent.
-                    </p>
-                  </li>
-                  <li>
-                    <dl class="switch">
-                      <dt>If <var>audio/video type</var> is Video:</dt>
-                      <dd>
+                    <p class="note">If <var>media types</var> is empty (i.e. only container is specified) then
+                        definate support implies support of all possible codecs and parameters for <var>container</var>
+                        which is unlikely for most container types.</p>
+                    <ol>
+                      <li>
                         <p>
-                          Add <var>requested media capability</var> to the <a def-id="option-videoCapabilities"></a> member of <var>local accumulated configuration</var>.
+                          Add <var>requested media capability</var> to <var>supported media capabilities</var>.
                         </p>
-                      </dd>
-                      <dt>If <var>audio/video type</var> is Audio:</dt>
-                      <dd>
-                        <p>
-                          Add <var>requested media capability</var> to the <a def-id="option-audioCapabilities"></a> member of <var>local accumulated configuration</var>.
+                        <p class="note">
+                          This step ensures that the values of the members of entries in <var>supported media capabilities</var> are exactly the strings supplied in
+                          <var>requested media capability</var> without modification by the User Agent.
                         </p>
-                      </dd>
-                    </dl>
-                    <p class="note">
-                      This step ensures that configurations are always checked with configurations from previous iterations, including from previous calls to this algorithm.
-                      Otherwise, only configurations from previous calls to this algorithm would be checked in subsequent calls.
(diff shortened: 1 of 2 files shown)


encrypted-media-respec.html
```diff
@@ -839,37 +839,42 @@
                     <p class="note">For example, if <var>audio/video type</var> is Video and the top-level type is not "video" or <var>media types</var> contains non-video codecs.</p>
                   </li>
                   <li><p>If <var>robustness</var> is not the empty string and contains an unrecognized value or a value not supported by <var>implementation</var>, continue to the next iteration. String comparison is case-sensitive.</p></li>
-                  <li><p>If the user agent and <var>implementation</var> do not support playback of encrypted <a def-id="media-data"></a> for the combination of <var>container</var>, <var>media types</var>, <var>robustness</var> and <var>local accumulated configuration</var>, continue to the next iteration.</p>
+                  <li><p>If the user agent and <var>implementation</var> definately support playback of encrypted <a def-id="media-data"></a> for the combination of <var>container</var>, <var>media types</var>, <var>robustness</var> and <var>local accumulated configuration</var>:</p>
```
ddorwin: definitely

encrypted-media-respec.html
```diff
@@ -839,37 +839,42 @@
...
-                      This step ensures that the values of the members of entries in <var>supported media capabilities</var> are exactly the strings supplied in
-                      <var>requested media capability</var> without modification by the User Agent.
-                    </p>
-                  </li>
-                  <li>
-                    <dl class="switch">
-                      <dt>If <var>audio/video type</var> is Video:</dt>
-                      <dd>
+                    <p class="note">If <var>media types</var> is empty (i.e. only container is specified) then
+                        definate support implies support of all possible codecs and parameters for <var>container</var>
```
ddorwin: definite
Also fix the commit message.

encrypted-media-respec.html
```diff
@@ -839,37 +839,42 @@
...
-                    <p class="note">
-                      This step ensures that the values of the members of entries in <var>supported media capabilities</var> are exactly the strings supplied in
-                      <var>requested media capability</var> without modification by the User Agent.
-                    </p>
-                  </li>
-                  <li>
-                    <dl class="switch">
-                      <dt>If <var>audio/video type</var> is Video:</dt>
-                      <dd>
+                    <p class="note">If <var>media types</var> is empty (i.e. only container is specified) then
```
ddorwin: nit: I think this is less ambiguous with the variable if "the" is added after "only".

encrypted-media-respec.html
ddorwin: suggestion: Emphasize all: <em>all</em>

ddorwin: Regarding "codecs and parameters": "codecs" is one of the parameters. RFC6381 (see line 833) defines "codecs" and "profiles".

I think the confusion may come from line 835: "the set of codecs, with parameters, specified by <var>parameters</var>" is confusing. This comes from https://github.com/w3c/encrypted-media/commit/59c28e9a5ab0f53ee6795bf3373f07852f4dd850, but I'm not sure what the intent is. Perhaps we should s/, with parameters,/ and other capabilities/. This would stress codecs while also handling profiles and anything else that might be added to line 833 in the future.

encrypted-media-respec.html
```diff
@@ -839,37 +839,42 @@
...
+                        definate support implies support of all possible codecs and parameters for <var>container</var>
+                        which is unlikely for most container types.</p>
```
ddorwin: Even if a user agent knew it supported all codecs supported by a container type _today_, that could change in the future. I suggest changing this to something like "unless the container explicitly implies codecs." Even then, we might be better off handling that above. See comment at line 844.

encrypted-media-respec.html
ddorwin: Line 835 says _parameters_ may be empty. Rather than just allowing that and trying to handle it in this Note, perhaps we should handle it there.

We have two options for handling empty.
1. Set _parameters_ to _all_ possible parameters for _container_.
2. If _container_ implies a set of codecs and other parameters, let _parameters_ be that set. Otherwise, continue to the next iteration.

I suggest that we choose (2). See the comment at line 846 for why.

ddorwin:
Thanks. The "definitely" part looks fine. Most of the comments are about handling container-only strings.

Also, please ensure the commit message has "definite" and "Fix #151: ..." before committing.

encrypted-media-respec.html
mwatson2: I agree with the above comments. I've updated the PR:

- to refer to "codecs and codec constraints" (Any information in addition to the codec name is essentially a constraint e.g. level and profile)
- adopt (2) above for the "container only" case

encrypted-media-respec.html
```diff
@@ -832,44 +832,65 @@
...
                   <li><p>If the user agent does not recognize one or more <var>parameters</var>, continue to the next iteration.</p></li>
-                  <li><p>Let <var>media types</var> be the set of codecs, with parameters, specified by <var>parameters</var>. It MAY be empty. The case-sensitivity of string comparisons is determined by the appropriate RFC or other specification.</p>
+                  <li><p>Let <var>media types</var> be the set of codecs and codec constraints specified by <var>parameters</var>. The case-sensitivity of string comparisons is determined by the appropriate RFC or other specification.</p>
                     <p class="note">Case-sensitive string comparison is RECOMMENDED because RFC 6381 [[RFC6381]] says, "Values are case sensitive" for some formats.</p>
                   </li>
+                  <li>
+                    <p>
+                      If <var>media types</var> is empty:
+                      <dl class="switch">
+                        <dt>If <var>container</var> implies a specific set of codecs and codec constraints:</dt>
```
ddorwin: I worry a little about the interpretation of "implies" (for example, an erroneous assumption that "video/mp4" implies AVC1 and AAC) and would like to be more explicit/specific, but nothing simple comes to mind.

ddorwin:
LGTM. Thanks

encrypted-media-respec.html
mwatson2: "normatively implies" ? At least that suggests that it needs to be a concrete specified thing, not just a common practice

ddorwin: Great!

mwatson2:
Merged as 2e5ab17704e19e3882aa2a7c9c0bc07b026c2897

Issue #57, opened 2015-05-18
Clarify behavior for denied user consent


ddorwin:
Step 15 of the _Get Supported Configuration_ algorithm defines when to prompt for use of Distinctive Identifier(s). (Issue #51 is related to cleaning this up.) However, it does not say what to do when such consent is denied.

The simple solution is to return null as in step 14.

However, it's possible that the configuration could have been supported in a different way^. We may want to allow reprocessing of the _candidate configuration_ (plus changes based on the denied consent).

We might also want to generalize the text, while still emphasizing the importance for Distinctive Identifiers, to cover checking consent for other reasons (beyond the existing note).

^ For example, if _candidate configuration_'s `distinctiveIdentifier` value is `"optional"`, knowing that the consent is denied might have allowed selection of a different set of `videoCapabilities` that don't require a Distinctive Identifier).

mwatson2:
When did we agree to "recommend" user consent prompts for Distinctive Identifiers ? I thought such prompts were an issue for user agents. Since the identifier is per-site, clearable and encrypted it differs from a cookie only in that it is not easy to steal. For cookies, depending on jurisdiction, it's common simply to provide information to users, rather than seek a priori consent. This should be left open for User Agents to work out the best path, rather than making a recommendation in the specification.

mwatson2:
I opened a new issue (https://github.com/w3c/encrypted-media/issues/63) for the recommendation related to user consent.

This issue can be unblocked to address the question of what should happen if consent is requested and denied. IIUC, one of the points of the site providing multiple options to the requestMediaKeySystemAccess was that it could deal with fallback to lower capabilities for the case where capabilities are unavailable or denies in a single step.

So, the procedures should ensure that when consent is denied, it behaves the same as if that configuration wasn't physically supported. Another requested configuration might still work.

ddorwin:
>IIUC, one of the points of the site providing multiple options to the requestMediaKeySystemAccess was that it could deal with fallback to lower capabilities for the case where capabilities are unavailable or denies in a single step.

The issue here is the combination of user permissions and `optional` along with the various other "optional" parts of the configuration (i.e. `videoCapabilities` entries).

[...]

Issue #96, opened 2015-09-24
Consider reasons other than Distinctive Identifier in the steps of the Consent Status algorithm


ddorwin:
PR #66 adds a "Get Distinctive Identifier Consent Status" algorithm that determines whether user consent is necessary. There are two areas for improvement:

1. The algorithm takes an <var>accumulated configuration</var> but doesn't use it.
2. The step being refactored contains a note that "Implementations MAY require consent for other reasons, such as the security properties of the CDM implementation." This could be made into a step that would fit better in the algorithm ([original suggestion](https://github.com/mwatson2/encrypted-media/commit/0d33734864311d6c61ad0f1af3f2213fd0611e39#commitcomment-13354947)).

Addressing (2) may use <var>accumulated configuration</var>.

I'll work on this after #66 is merged.

mwatson2:
This issue is addressed in the PR for issue #57 (PR #165)

mwatson2:
This is expected to be solved by #57

mwatson2:
#57 is now closed.

PR #165, opened 2016-04-14
PR for issue #57: Repeat configuration selection on denial of user co…

mwatson2:
…nsent for distintive identifiers


encrypted-media-respec.html
```diff
@@ -606,7 +606,83 @@
...
             <h5>Get Supported Configuration</h5>
             <p>Given a <a def-id="keysystems"></a> implementation <var>implementation</var>, <a>MediaKeySystemConfiguration</a> <var>candidate configuration</var>, and <var>origin</var>, this algorithm returns a supported configuration or <code>null</code> as appropriate.</p>
-            <p class="note">Unrecognized dictionary members in <var>candidate configuration</var> are ignored per [[WebIDL]] and will never reach this algorithm. Thus, they cannot be considered as part of the configuration.</p>
+            <p class="note">Unrecognized dictionary members in <var>candidate configuration</var> are ignored per [[WebIDL]] and will never reach this algorithm. Thus, they cannot be considered as part of the configuration.
+            </p>
+            <ol>
+              <li>
+                <dl class="switch">
+                  <dt>
+                    If the <a def-id="option-distinctiveIdentifier"></a> member of
```
ddorwin: I worry about using `distinctiveIdentifier` as the (only) variant. For example, in #96, we want to account for other potential reasons. Perhaps we should use something like "has consent been denied", "has consent been requested", or "can request consent".

Then, anywhere that consent might be required (including by choice of the user agent), we could check this value and choose not to do whatever would require consent.

I don't think this materially changes the behavior. Passing a separate Boolean also allows us to avoid copying and modifying the configuration.

ddorwin: I think we might also end up with more common steps here. For example, "If _supported configuration_ is `ConsentDenied`, &lt;run the Get Supported Configuration and Consent algorithm with the opposite value&gt;."

encrypted-media-respec.html
```diff
@@ -606,7 +606,83 @@
...
+                      </li>
+                    </ol>
+                  </dd>
+                </dl>
+              </li>
+            </ol>
+          </section>
+          <section>
+            <h5>Get Supported Configuration and Consent</h5>
+            <p>Given a <a def-id="keysystems"></a> implementation <var>implementation</var>, <a>MediaKeySystemConfiguration</a> <var>candidate configuration</var>, and <var>origin</var>, this algorithm returns a supported configuration, <code>null</code>, or <code>ConsentDenied</code> as appropriate.
```
ddorwin: Although this allows the call sites to be shorter (there is only one condition to handle because null and valid configurations both get forwarded), I wonder if we should be consistent in how we handle the unsuccessful cases by returning `NotSupported` rather than `null`. That would make _this_ algorithm easier to read. It might also make it more clear what is happening at the call sites and that `null` is a possible value when we "return _supported configuration_."

encrypted-media-respec.html
```diff
@@ -606,7 +606,83 @@
...
+                  <dt>
+                    If the <a def-id="option-distinctiveIdentifier"></a> member of
+                    <var>candidate configuration</var> is <a def-id="requirement-optional"></a>:
+                  </dt>
+                  <dd>
+                    <ol>
+                      <li>
+                        <p>
+                          For each <var>distinctive identifier state</var> in the list
+                          [ <a def-id="requirement-optional"></a>, <a def-id="requirement-not-allowed"></a> ]:
```
ddorwin: In addition to my comments above, we need to remember that consent was requested for future configurations. If consent was denied for this configuration, we need to reevaluate it (as this PR currently does) AND if this configuration is not supported (this algorithm terminates with `null`), evaluate all remaining configurations with consent denied.

That's actually covered by your update to the Note below. Thus, algorithmically, maybe we are fine. I like that implementations have flexibility to request multiple consents (or equivalent) if necessary. It would be nice if this was more clear, though. Perhaps we can add another Note or move that part of the Note up here or in `requestMediaKeySystemAccess()` to make that clearer to the reader.

ddorwin:
Thanks. The approach looks good. I have some comments above on the details.

encrypted-media-respec.html
mwatson2: Do you mean to say we should completely generalize the consent mechanism ? Presently we talk explicitly about "consent for use of distinctive identifiers with _configuration_". Since the distinctive identifier state is part of _configuration_ we could generalize and simply talk about "consent to use _configuration_" and then in the leaf "get consent status" algorithm we would implement the one normative requirement (that distinctive identifier consent is required unless all the recommendations of the security section are followed).

I would also add a note about User Agents having flexibility to determine the set of _configuration_ values that a given consent applies to, since this depends on the language that was used to obtain the consent and UA decisions on how granular to be about asking for consent for different things.

We then still have to deal with the "optional" cases. We could punt this to the application (remove the `optional` states and force the application to list the alternatives it would accept. But the `optional` states do avoid having to deal with a combinatoric explosion at the application: instead we end up having to try every combination in our algorithm to find one that works.

mwatson2:
Ok, so I have updated this Pull request as follows:
- use explicit `NotSupported` return value instead of null
- add explanation of the iterative consent process at the start of Get Supported Configurations
- generalize the iteration to try all combinations of any number of optional MediaKeyRequirement members (we only have two, so the generalization is perhaps overkill, but it would work unchanged if a third MediaKeyRequirement member was added in future).
- generalize the consent algorithm for any kind of consent (not just distinctive identifiers) that the UA requires (the spec only requires consent for distinctive identifiers if the recommendations of the security section are not implemented, but UAs may require consent for more).

mwatson2:
I believe this PR now addresses issue #96 as well.

encrypted-media-respec.html
```diff
@@ -605,8 +605,112 @@
...
+            <div class="note">
+              <p>
+              For certain configurations, it may be required to obtain user consent or inform the user. User Agents have some flexibility to determine
+              whether consent is required for a specific configuration and whether such consent may also apply to other configurations. For example,
+              consent to one configuration may also imply consent for less powerful, more restricted configurations. Equally, a denial of consent for
+              one configuration may imply denial of consent for more powerful, less restricted configurations.
+              </p>
+              <p>
+              Supported configurations, including supported audio and video codecs, may depend on availability of optional capabilities such as
+              <a def-id="distinctive-identifier"></a>s and persistent state. The following algorithm iteratively tries configurations with different
```
ddorwin: You can link the entire word, including the 's', by putting "Distinctive Identifiers" inside the element.

In this case, though, 'distinctive-identifiers' already exists.

encrypted-media-respec.html
```diff
@@ -605,8 +605,112 @@
...
+              requirement values for these optional configuration elements until consent is achieved.
```
ddorwin: It's not necessarily that consent is achieved. It's possible that it falls back to a configuration that does not require consent (or fails).

encrypted-media-respec.html
```diff
@@ -605,8 +605,112 @@
...
+              </p>
+              <p>
+              User Agents should persist consent responses at least for the duration of the <a def-id="requestMediaKeySystemAccess"></a>
```
ddorwin: s/should/MAY/?

Should we use a word with less baggage than "persist?" For example, s/should persist consent responses at least/MAY {remember|reuse} consent/.

mwatson2: I think this one should be "should" because if you don't (persist for at least the getMediaKeySystemAccess) then you'll have many user consent prompts. I agree about "persist".

ddorwin: It's a minor issue, but I was trying to convey "implementations probably want to/most likely will do this" without implying "this is the only way."

OTOH, perhaps this or similar _should_ be recommended because authors will write apps assuming there will be at most one prompt.

encrypted-media-respec.html
```diff
@@ -605,8 +605,112 @@
...
+              algorithm in order to avoid repeated requests to the user for similar configurations.
+              </p>
+            </div>
+            <ol>
+              <li>
+                <p>Let <var>requirements</var> be the list [ <code>distinctiveIdentifier</code>, <code>persistentState</code> ].</p>
```
ddorwin: I wonder whether we need to loop through all the combinations and whether this actually gives the desired behavior. Is there a reason that looping twice with "behave as if you could request consent" and (if no consent) "what can you do without consent" (the difference in which is captured in a Boolean).

On a related note, the order that the requirements are checked in steps 13 and 14 is intentional. I think you preserve this below, but it is sort of weird and not guaranteed to be as desired if another requirement was added.

encrypted-media-respec.html
```diff
@@ -605,8 +605,112 @@
...
+                </p>
+              </li>
+              <li>
+                <dl class="switch">
+                  <dt>If <var>optional requirements</var> is not empty:</dt>
+                  <dd>
+                    <ol>
+                      <li>
+                        <p>
+                          Let <var>optional requirement combinations</var> be the list of all sublists of <var>optional requirements</var>
```
ddorwin: This is difficult to understand, and use of length/position seems unnatural vs. alphabetical. (Am I correct about the reason for this above?)

encrypted-media-respec.html
mwatson2: Presumably, user consent may depend on any or all of the MediaKeySystemRequirement elements (i.e. the user may give consent for distinctive identifiers but not for persistent store or vice versa) and indeed the rest of the _accumulated configyration_. In practice, I would expect User Agents to bundle these into one question, or perhaps use some check-boxes, but how it is presented is up to the UA.

When a site marks a capability as optional it is basically requesting that capability (if needed for the requested media capabilities) but also asking that the UA fall back to whatever can be done without that capability. We could remove the optional value altogether and just have the site specify what it would accept as fallback, but that means the site listing a lot of alternatives.

If we're going to ask the UA to do the fallback, I think it basically has to try every combination.

encrypted-media-respec.html
mwatson2: Yes, this is what I meant by the generic version perhaps being overkill. Since we only have two MediaKeyRequirement there are only 4 possibilities, so we could list them explicitly. If we add another one we need the generic algorithm.

The ordering is important. We are ordering a list of lists, where the elements in each list are the ones which will be left as "optional". So we start leaving them all as optional (the longest list = the strongest requirement). If that fails, we try with exactly one of the optional items marked down to not-allowed. The order in which we try marking them down is the order we give for them at the start (in _requirements_).

At the moment we will try { distintiveIdentifier: not-allowed, persistentStore: optional } before we try { distintiveIdentifier: optional, persistentStore: not-allowed }. We could do it the other way around.

I'm not sure there is a definitive way to determine what is the best order: the important thing will be to have the order consistent across UAs. If the site wants to test these in a different order it can list them explicitly, without using the "optional" value.

Anyway, I'm happy to use a simpler algorithm. Or I could spell out the steps here in more detail and provide an example.

ddorwin: I'm not sure why we have a list of lists. Specifically, why is there a "list of all sublists of <var>optional requirements</var>?" Is this a list of all possible combinations?

It might help to have examples of some of these variables for discussion. (We can then add them as Notes if necessary.

mwatson2: Yes, a list of all possible combinations:
```
{ distinctiveIdentifier: optional, persistentState: optional }
{ distinctiveIdentifier: not-allowed, persistentState: optional }
{ distinctiveIdentifier: optional, persistentState: not-allowed }
{ distinctiveIdentifier: not-allowed, persistentState: not-allowed }
```
(Where each of those combinations is represented not by an object as shown about but by a list of the optional items).

Now, I could list those four explicitly and have a step to remove the ones that don't apply (based on the input values of `persistentState` and `distinctiveIdentifier`) but that wouldn't scale to a third MediaKeyRequirement.

I can certainly add examples. I'm still wondering if the whole thing is overkill. WDYT ?

encrypted-media-respec.html
ddorwin: I'm still thinking about this, but I'll post my current thoughts for discussion.

<br/>One thing I forgot to mention: It seems unlikely that a "not-allowed" value would require consent if a "required" value does not. As a bit of a corollary, I don't think denied consent should enable alternate options. (These are more gut instinct than definitive statements.)

Thus, do we need to run through all the combinations (or can we skip at least some based on denied consent)? (It's possible the observable behavior of the current text in the PR is the same, though it's hard to verify this currently.)


<br/>Thinking through possible values and implementations:

The "worst case" for the algorithm is that both are "optional". In that case, an implementation might choose [something that results in] "required" for DI and prompt. (PS could be either value.) If the consent is declined, that almost certainly means that DI should not be "not-allowed".

However, let's say the implementation chose [something that results in] "required" for PS too and asked for consent to both DI and PS (and maybe something else), or even just PS. In that case, it is _theoretically_ possible that the implementation could decide that PS should be "not-allowed" and it should retry with DI as "optional".

In the next loop, the implementation can choose [functionality that results in] any value for DI. In the worst case, it picks "required" again and prompts again. In another scenario, DI might only work with PS, resulting in the selection of "not-allowed". (I can't see how this could happen given the explicit separation of the two types of "persistence.") Perhaps the implementation still wants to prompt even though both DI and PS are "not-allowed".

All these combinations are possible if we iterate through all possible combinations. However, the order in which they can be tried is also prescribed, preventing the user agent from trying in its preferred order.


<br/>Perhaps most importantly, relying solely on varying the `MediaKeysRequirement` values also precludes varying other implementation-specific properties that could require consent.


<br/>Perhaps we can simplify this by allowing implementations to request consent only once per configuration. (As you said, how this is handled is up to the implementation.) Specifically:

1. Let the current algorithm run once with all the existing handling of "optional", order of importance, etc.
1. If consent is denied, let _restrictions_ = the set of things not consented to.
1. Update the "optional" requirements based on _restrictions_.
1. Rerun the algorithm exactly once with the new values.

This doesn't handle lack of consent for other things. To address that, we could pass a "_restrictions_ object" (empty in the first run) and check it along with _accumulated configuration_ in each such step (i.e. "...in combination with _accumulated configuration_ **and _restrictions_**").

While this _may_ be more limiting, it is clearer and more descriptive of how we intend for things to work (it's hard to verify that the current normative text in this PR has the same or similarly desired outcome) and allows consideration of implementation-specific properties. Also, the counterexamples for multiple rounds or multiple prompts seem unlikely, though we should acknowledge that we can't predict all implementations, especially those that might have multiple pipelines.

mwatson2: > The "worst case" for the algorithm is that both are "optional". In that case, an implementation might choose [something that results in] "required" for DI and prompt. (PS could be either value.) If the consent is declined, that almost certainly means that DI should not be "not-allowed".

Do you mean *should* be "not-allowed" ? (Consent was denied).

> Perhaps we can simplify this by allowing implementations to request consent only once per configuration. (As you said, how this is handled is up to the implementation.)

Certainly we can simplify by *requiring* implementations to request consent only once. The only problem is that the consent is to the use (or not) of DI and PS *in association with _accumulated configuration_*. If all the subsequent configurations tried are subsets of _accumulated configuration_ (from a consent perspective) we are ok: the original consent still applies. But if not, we are stuck.

For example, the site marks both DI and PS as optional. The implementation chooses codecs and other options. that require both. The user consents only to DI "in association with those codecs and other options". So, we try again with DI required and PS not allowed. The set of codecs and other options now chosen is different. Does the prior consent still apply ? Perhaps there is a robustness level that implies hardware DRM that works without PS but that if you want to use a lower level (sw) that requires PS and perhaps the UA distinguishes HW and SW from a consent perspective ?

encrypted-media-respec.html
```diff
@@ -605,8 +605,112 @@
...
+                          <li>
+                            <p>
+                            Let <var>supported configuration</var> be the result of executing the
+                            <a def-id="get-supported-configuration-and-consent-algorithm"></a> algorithm
+                            with <var>implementation</var>, <var>modified candidate configuration</var> and <var>origin</var>.
+                            </p>
+                          </li>
+                          <li>
+                            <p>
+                            If <var>supported configuration</var> is not <code>ConsentDenied</code>,
```
ddorwin: Based on this step, wouldn't `NotSupported` from one of the combinations cause us to not check the others? Might just need to add "or ` <code>NotSupported</code>`".

encrypted-media.js
```diff
@@ -172,8 +172,9 @@
...
     'requestMediaKeySystemAccess-call': { func: idlref_helper, fragment: 'widl-Navigator-requestMediaKeySystemAccess-Promise-MediaKeySystemAccess--DOMString-keySystem-sequence-MediaKeySystemConfiguration--supportedConfigurations', link_text: 'requestMediaKeySystemAccess',  },
-    'get-distinctive-identifier-consent-status-algorithm' : { func: term_helper, fragment: 'get-distinctive-identifier-consent-status', link_text: 'Get Distinctive Identifier Consent Status' },
+    'get-consent-status-algorithm' : { func: term_helper, fragment: 'get-consent-status', link_text: 'Get Consent Status' },
     'get-supported-configuration-algorithm': { func: term_helper, fragment: 'get-supported-configuration', link_text: 'Get Supported Configuration',  },
+    'get-supported-configuration-and-consent-algorithm': { func: term_helper, fragment: 'get-supported-configuration', link_text: 'Get Supported Configuration and Consent',  },
```
ddorwin: Add `-and-consent` to the fragment.

encrypted-media-respec.html
mwatson2: I _think_ that if a configuration is not supported with some MediaKeyRequirement fields set to "optional" then it is also not supported with those same fields downgraded to "not-allowed", since the check for support considers both with / without the optional item. Only if consent is denied is it worth trying the downgrade, because the original check for support might have returned something that requires the optional item and then consent may have been denied for that.

ddorwin: In theory, the user agent could have made choices that require two "optional" `MediaKeysRequirement` items to be "required". If consent was denied for reasons that aren't the first item and "not-allowed" is not supported for that item, this could break. Yes, this is unlikely, especially given these two items, but it seems that if the algorithm is going to be very general, it should not make such assumptions.

An example:
>Trying ["optional", "optional"] => becomes ["required", "required"] => `ConsentDenied` => ... trying ["not-allowed", "optional"] => `NotSupported`

In theory, this could continue:
>=> trying ["optional", "not-allowed"] => becomes ["required", "not-allowed"] => Supported

encrypted-media-respec.html
ddorwin: Thanks, that is helpful. If we stick with this solution, we can try to determine the best way to document this.

To summarize the current approach:
* This iterates through all these combinations until the result is not `ConsentDenied`.
* If consent was denied the first time, one (or more) of the "not-allowed" values may result in consent granted or consent not being required. (See the comment below about potentially failing too early.)
* This relies on "brute force" and does not specifically consider the consent request or details of the denial.
* This does not handle _optional_ implementation-specific consent-requiring properties unrelated to a `MediaKeysRequirement`.

encrypted-media-respec.html
ddorwin: >> The "worst case" for the algorithm is that both are "optional". In that case, an implementation might choose [something that results in] "required" for DI and prompt. (PS could be either value.) If the consent is declined, that almost certainly means that DI should not be "not-allowed".

>Do you mean *should* be "not-allowed" ? (Consent was denied).

Yes. I corrected my comment with strikethrough to help any future readers.


<br/> Good point about the configuration relationships. I think we can address this by replacing step 4 with rerunning until there are no other possible configurations. I think we might be able to express that with `NotSupported` since we are expecting the implementation to pick supported values rather than trying possible `MediaKeysRequirement` values.

What do you think about that approach in general?

mwatson2: Yes, that's better.

To check we're on the same page, the changes will be:
- remove the construction of "all possible combinations"
- change the loop to "repeat until done" in some form
- Get Supported Configurations and Consent will return a _restrictions_ list with the `ConsentDenied` return value. The _restrictions_ list is a list of those elements of the _modified configuration_ of type `MediaKeyRequirement` which had value `optional` and for which consent has been denied.
- If _restrictions_ is non-empty, update _modified candidate configuration_ to set all the members listed in _restrictions_ to `not-allowed` and continue the loop
- If _restrictions_ is empty, it means none of the `MediaKeyRequirement` elements were optional and consent was denied anyway, so we return `NotSupported`.

However, we might have a different problem, unrelated to the handling of these `optional` values that could affect this section: suppose the site sets DI to `required` and asks for some set of codecs with hardware DRM or a different set of codecs with software DRM. And now suppose the software DI implementation supports all the recommendations of the security section (and so requires the user to be informed, but not consent) but the HW one does not. The user denies consent for DI with HW DRM, but the SW option is still viable. I don't believe we can handle this with the existing text.

encrypted-media-respec.html
mwatson2: Yep, I think that's accurate and perhaps my DI example above is an example of the last bullet.

encrypted-media-respec.html
mwatson2: Yes, you're right.

encrypted-media-respec.html
ddorwin: I forgot to reply to this and we are already going down this path, but for the record: I think I agree with the first two paragraphs but think we should keep "optional" for the reasons you mention.

encrypted-media-respec.html
ddorwin: >I don't believe we can handle this with the existing text.

To make sure I understand, is this because DI is set to `not-allowed` based on the _restrictions_ list?


<br/>Thanks for listing the bullets. My idea as slightly different (see the paragraph above related to "handl[ing] lack of consent for other things"), and I think it might address this problem. Thus, the changes will be:
- remove the construction of "all possible combinations"
- change the loop to "repeat until done"**, which is probably until the result is not `ConsentDenied`**
- Get Supported Configurations and Consent will return a _restrictions_ list with the `ConsentDenied` return value. The _restrictions_ list is a list of those **properties ~~elements of the _modified configuration_ of type `MediaKeyRequirement` which had value `optional` and~~** for which consent has been denied. **The contents of _restrictions_ is implementation-specific, but it likely has at least one property related to DI.**
- **~~If _restrictions_ is non-empty, update _modified candidate configuration_ to set all the members listed in _restrictions_ to `not-allowed` and continue the loop~~ The _restrictions_ list is passed to the method, which is updated to consider it wherever (support in combination with) _accumulated configuration_ is considered. The value of the "optional" values is not changed, but it is likely that "required" is no longer a valid value when _restrictions_ is considered.**
- **~~If _restrictions_ is empty, it means none of the `MediaKeyRequirement` elements were optional and consent was denied anyway, so we return `NotSupported`.~~ We continue to return `NotSupported` whenever we reach a point where there are no supported configurations, including due to the application of _restrictions_. (_restrictions_ is only returned/considered when `ConsentDenied` is returned.) This allows non-consent-related failures to work as before while also addressing the loop termination.**

encrypted-media-respec.html
```diff
@@ -605,8 +605,56 @@
...
+            <div class="note">
+              <p>
+              For certain configurations, it may be required to obtain user consent or inform the user. User Agents have some flexibility to determine
+              whether consent is required for a specific configuration and whether such consent may also apply to other configurations. For example,
+              consent to one configuration may also imply consent for less powerful, more restricted configurations. Equally, a denial of consent for
+              one configuration may imply denial of consent for more powerful, less restricted configurations.
+              </p>
+              <p>
+              Supported configurations, including supported audio and video codecs, may depend on availability of optional capabilities such as
+              <a def-id="distinctive-identifier"></a>s and persistent state. The following algorithm iteratively tries to find a configuration
```
ddorwin: `<a def-id="distinctive-identifiers">` then remove the extra 's'.

ddorwin: Actually, `<a def-id="distinctive-identifier">Distinctive Identifier(s)</a>` might read better here.

encrypted-media-respec.html
```diff
@@ -605,8 +605,56 @@
...
+              that is both supported and has user consent, each iteration considering a stronger set of restrictions derived from user consent prompts.
```
ddorwin: s/prompts/checks/?
We don't expect a series of prompts.

ddorwin: I understand "stronger set of restrictions", but it seems a bit ominous. Adding "... on the configuration" might help. Or something like "placing stronger restrictions on the configuration."

encrypted-media-respec.html
```diff
@@ -605,8 +605,56 @@
...
+                </li>
+              </li>
+            </ol>
+          </section>
+          <section>
+            <h5>Get Supported Configuration and Consent</h5>
+            <p>Given a <a def-id="keysystems"></a> implementation <var>implementation</var>, <a>MediaKeySystemConfiguration</a> <var>candidate configuration</var>,
+              <var>restrictions</var> and <var>origin</var>, this algorithm returns a supported configuration, <code>NotSupported</code>, or <code>ConsentDenied</code> as appropriate and, in the <code>ConsentDenied</code> case, <var>restrictions</var>.
+            </p>
+            <p class="note"><var>restrictions</var> consists of an implementation-specific set of configurations for which consent has been denied by the user.</p>
```
ddorwin: "configurations" could be confusing here since it could be confused with a `MediaKeySystemConfiguration`.
"Properties" might work but is perhaps too vague. "Configuration options?" "Implementation configuration properties?"

encrypted-media-respec.html
```diff
@@ -651,15 +699,15 @@
...
-                    <p>If the <var>implementation</var> does not support a persistent <a def-id="distinctive-identifier"></a> in combination with <var>accumulated configuration</var>, return <code>null</code>.</p>
+                    <p>If the <var>implementation</var> does not support a persistent <a def-id="distinctive-identifier"></a> in combination with <var>accumulated configuration</var> and <var>restrictions</var>, return <code>NotSupported</code>.</p>
                   </dd>
                   <dt><a def-id="requirement-optional"></a></dt>
                   <dd>
                     <p>Continue.<!-- Will be updated below. --></p>
                   </dd>
                   <dt><a def-id="requirement-not-allowed"></a></dt>
                   <dd>
-                    <p>If the <var>implementation</var> requires a <a def-id="distinctive-identifier"></a> in combination with <var>accumulated configuration</var>, return <code>null</code>.</p>
```
ddorwin: Unrelated to this CL, we should probably replace "Continue" with "Continue to the next step" to avoid any potential confusion with `continue` in programming languages.

encrypted-media-respec.html
```diff
@@ -699,8 +747,8 @@
...
-                    <p>Unsupported combinations include but are not limited to a <var>session type</var> for which the <a def-id="is-persistent-session-type-algorithm"></a> algorithm returns <code>true</code> when <var>accumulated configuration</var>'s <a def-id="option-persistentState"></a> value is <a def-id="requirement-not-allowed"></a>.</p>
+                    <p>If the <var>implementation</var> does not support <var>session type</var> in combination with <var>accumulated configuration</var> and <var>restrictions</var>, return <code>NotSupported</code>.</p>
+                    <p>Unsupported combinations include but are not limited to a <var>session type</var> for which the <a def-id="is-persistent-session-type-algorithm"></a> algorithm returns <code>true</code> when <var>accumulated configuration</var>'s <a def-id="option-persistentState"></a> value is <a def-id="requirement-not-allowed"></a> or <var>accumulated configuration</var>'s <a def-id="option-persistentState"></a> value is <a def-id="requirement-optional"></a> but persistent state is disallowed by <var>restrictions</var>.</p>
```
ddorwin: Should we just address this (and DI) in the "optional" cases above? For example, replace "Continue" with "If _restrictions_ prevents..., set the value to `"not-allowed"`." That seems more explicit and maintainable.

encrypted-media-respec.html
```diff
@@ -709,16 +757,16 @@
 
               <li><p>If the <a def-id="option-videoCapabilities"></a> member is <a def-id="present-dictionary-member"></a> in <var>candidate configuration</var>:</p>
                 <ol>
-                  <li><p>Let <var>video capabilities</var> be the result of executing the <a def-id="get-supported-capabilities-for-audio-video-type-algorithm"></a> algorithm on Video, <var>candidate configuration</var>'s <a def-id="option-videoCapabilities"></a> member, and <var>accumulated configuration</var>.</p></li>
-                  <li><p>If <var>video capabilities</var> is <code>null</code>, return <code>null</code>.</p></li><!-- Video capabilities were specified, but none were supported. -->
+                  <li><p>Let <var>video capabilities</var> be the result of executing the <a def-id="get-supported-capabilities-for-audio-video-type-algorithm"></a> algorithm on Video, <var>candidate configuration</var>'s <a def-id="option-videoCapabilities"></a> member, <var>accumulated configuration</var> and <var>restrictions</var>.</p></li>
```
ddorwin: nit: For consistency, put a comma before "and" here and below.

encrypted-media-respec.html
```diff
@@ -750,73 +798,63 @@
...
+              <li><p>If <var>implementation</var> in the configuration specified by the combination of the values in <var>accumulated configuration</var> is not supported or not allowed in the <var>origin</var>, return <code>NotSupported</code>.</p>
                 <p class="note">In this step, "supported" includes the implementation being available for use when this algorithm returns, not just user agent support for such an implementation.</p>
               </li>
 
-              <li><p>If <var>accumulated configuration</var>'s <a def-id="option-distinctiveIdentifier"></a> value is <a def-id="requirement-required"></a>:</p>
+              <li>
+                <p>
+                  If <var>accumulated configuration</var>'s <a def-id="option-distinctiveIdentifier"></a> value is <a def-id="requirement-required"></a> and if
+                  the <a def-id="distinctive-identifier"></a> is not <a href="#per-origin-identifiers">unique per-origin</a> and
```
ddorwin: This should probably be something like "`...the **implementation's** <a def-id="distinctive-identifier"></a> is not...`" or "`<a def-id="distinctive-identifier">Distinctive Identifier(s)</a> associated with <var>accumulated configuration</var> are not...`"

encrypted-media-respec.html
```diff
@@ -750,73 +798,63 @@
...
+                  <a href="#allow-identifiers-cleared">clearable</a>:
+                </p>
                 <ol>
                   <li>
+                    <p>Add the configurations described by <var>accumulated configuration</var> to <var>restrictions</var>.</p>
```
ddorwin: Do we really want to add all the configurations? That includes specific content type choices, for example.

This one is a bit unique - it was a property of the DI implementation that means consent was denied. However, that technically doesn't even mean that DI(s) are not allowed - it's theoretically possible that another configuration would result in a compliant DI.

Thus, perhaps we should just return `ConsentDenied` without modifying restrictions.

encrypted-media-respec.html
ddorwin: Or "Continue with the following steps" as used twice below.

However, below I propose replacing both of these with an "If" statement, which would make this moot.

encrypted-media-respec.html
```diff
@@ -750,73 +798,63 @@
...
+                      member of <var>accumulated configuration</var> is <a def-id="requirement-required"></a>. Continue with the following steps.
                     </p>
-                  </li>
-                  <li>
+                  </dd>
+                  <dt><code>Allowed</code>:</dt>
+                  <dd>
                     <p>
-                      Let <var>distinctive identifier consent status</var> be the result of running the <a def-id="get-distinctive-identifier-consent-status-algorithm"></a> algorithm on <var>accumulated configuration</var> and <var>origin</var> and follow the steps for the value of <var>distinctive identifier consent status</var> from the following list:":
+                      Continue with the following steps.
```
ddorwin: Nit: There is only one more step. Same above.

encrypted-media-respec.html
```diff
@@ -750,73 +798,63 @@
...
 
-              <li><p>Return <var>accumulated configuration</var>.</p></li>
+              <li><p>Return <var>accumulated configuration</var></p></li>
             </ol>
           </section>
 
           <section>
             <h5>Get Supported Capabilities for Audio/Video Type</h5>
-            <p>Given an <var>audio/video type</var>, <a>MediaKeySystemMediaCapability</a> sequence <var>requested media capabilities</var>, and <a>MediaKeySystemConfiguration</a> <var>partial configuration</var>, this algorithm returns a sequence of supported <a>MediaKeySystemMediaCapability</a> values for this audio/video type or <code>null</code> as appropriate.</p>
+            <p>Given an <var>audio/video type</var>, <a>MediaKeySystemMediaCapability</a> sequence <var>requested media capabilities</var>, <a>MediaKeySystemConfiguration</a> <var>partial configuration</var> and <var>restrictions</var>, this algorithm returns a sequence of supported <a>MediaKeySystemMediaCapability</a> values for this audio/video type or <code>null</code> as appropriate.</p>
```
ddorwin: nit: comma before "and"

encrypted-media-respec.html
```diff
@@ -839,7 +877,8 @@
                     <p class="note">For example, if <var>audio/video type</var> is Video and the top-level type is not "video" or <var>media types</var> contains non-video codecs.</p>
                   </li>
                   <li><p>If <var>robustness</var> is not the empty string and contains an unrecognized value or a value not supported by <var>implementation</var>, continue to the next iteration. String comparison is case-sensitive.</p></li>
-                  <li><p>If the user agent and <var>implementation</var> do not support playback of encrypted <a def-id="media-data"></a> for the combination of <var>container</var>, <var>media types</var>, <var>robustness</var> and <var>local accumulated configuration</var>, continue to the next iteration.</p>
+                  <li><p>If the user agent and <var>implementation</var> do not support playback of encrypted <a def-id="media-data"></a> for the combination of <var>container</var>, <var>media types</var>, <var>robustness</var>, <var>local accumulated configuration</var> and <var>restrictions</var>,
```
ddorwin: nit: Should this be "with _restrictions_"?

Otherwise: nit: comma

FYI, I think you'll have a rebase conflict on this line.

encrypted-media-respec.html
```diff
@@ -887,42 +926,84 @@
           </section>
 
           <section>
-            <h5>Get Distinctive Identifier Consent Status</h5>
+            <h5>Get Consent Status</h5>
             <p>
-              Given an <var>accumulated configuration</var> and an <var>origin</var>, this algorithm returns the consent status of <a def-id="distinctive-identifier">Distinctive Identifier(s)</a> for <var>accumulated configuration</var> and <var>origin</var> as one of <code>ConsentRequired</code>, <code>InformUser</code>, <code>Allowed</code> or <code>Denied</code>.
+              Given an <var>accumulated configuration</var>, <var>restrictions</var> and an <var>origin</var>, this algorithm returns the consent status for
```
ddorwin: nit: comma
nit: remove second "an" (existing text)

encrypted-media-respec.html
```diff
@@ -605,8 +605,56 @@
...
+              consent to one configuration may also imply consent for less powerful, more restricted configurations. Equally, a denial of consent for
+              one configuration may imply denial of consent for more powerful, less restricted configurations.
+              </p>
+              <p>
+              Supported configurations, including supported audio and video codecs, may depend on availability of optional capabilities such as
+              <a def-id="distinctive-identifier"></a>s and persistent state. The following algorithm iteratively tries to find a configuration
+              that is both supported and has user consent, each iteration considering a stronger set of restrictions derived from user consent prompts.
+              </p>
+              <p>
+              User Agents should persist consent responses at least for the duration of the <a def-id="requestMediaKeySystemAccess"></a>
```
ddorwin: I think we agreed we should use a word with less baggage than "persist" here. Remember? Reuse?

encrypted-media-respec.html
```diff
@@ -887,42 +926,84 @@
...
             </p>
             <ol>
               <li>
                 <p>
-                  If there is persisted denial for <var>origin</var> indicating that <a def-id="distinctive-identifier">Distinctive Identifier(s)</a> are not allowed in conjunction with <var>accumulated configuration</var>, return <code>Denied</code>.
+                  If there is persisted denial for <var>origin</var> indicating that <var>accumulated configuration</var> is not allowed:
                 </p>
+                <ol>
+                  <li>
+                    <p>Add all the configurations for which consent has been denied to <var>restrictions</var>.</p>
```
ddorwin: Same comment as line 657 about "configurations."

encrypted-media-respec.html
```diff
@@ -887,42 +926,84 @@
...
               </li>
               <li>
                 <p>
-                  If any of the recommendations of <a href="#allow-identifiers-cleared">Allow Identifiers to be Cleared</a> are not supported by the User Agent, return <code>ConsentRequired</code>.
+                  If any of the following are true:
                 </p>
+                <ul>
+                  <li>
+                    <p>
+                      the <a def-id="option-distinctiveIdentifier"></a> member of <var>accumulated configuration</var> is not
```
ddorwin: nit: I think we generally start bullets with capital letters.

encrypted-media-respec.html
```diff
@@ -887,42 +926,84 @@
...
+                      <a def-id="requirement-not-allowed"></a> and any of the recommendations of
+                      <a href="#allow-identifiers-cleared">Allow Identifiers to be Cleared</a> are not supported by the
+                      User Agent.
```
ddorwin: I think this should be "implementation" since it also includes the CDM. (It appears we use these words somewhat interchangeably and inconsistently, but here we should cover both the UA and CDM.)

encrypted-media-respec.html
```diff
@@ -887,42 +926,84 @@
...
+                    </p>
+                  </li>
+                </ul>
+                <p>then:</p>
+                <ol>
+                  <li>
+                    <p>
+                      Request user consent to use <var>accumulated configuration</var> in the <var>origin</var> and wait for the user response.
+                    </p>
+                    <p class="note">User consent to use<var>accumulated configuration</var> is dependent at least on the value of the <a def-id="option-distinctiveIdentifier"></a> member.</p>
```
ddorwin: This appears to be new, and it's not clear to me what the intent is.

There is a missing space after "use".

encrypted-media-respec.html
```diff
@@ -887,42 +926,84 @@
...
+                  <li>
+                    <p>
+                      the user agent requires explicit user consent for the <var>accumulated configuration</var> for other reasons.
+                    </p>
+                    <p class="note">
+                      Another reason for requiring explicit user consent may be due to the security properties of the CDM implementation.
+                    </p>
+                  </li>
+                </ul>
+                <p>then:</p>
```
ddorwin: "Run the following steps:"

encrypted-media-respec.html
mwatson2: I'm not convinced that it isn't actually a set of `MediaKeySystemConfiguration` values, or at least equivalent to that: at the end of the day, the site supplies a (description of a) list of `MediaKeySystemConfiguration` values that it would accept and each consent prompt reduces the set of all possible allowed `MediaKeySystemConfiguration` values. We return the first item in the intersection of the site's list with that set.

What could be in _restrictions_ that isn't essentially an implementation-specific rule - or the input to a rule - that can determine whether a specific `MediaKeySystemConfiguration` is allowed or not ? Such a rule is logically equivalent to the set of `MediaKeySystemConfiguration` values for which the rule returns false.

encrypted-media-respec.html
```diff
@@ -887,42 +926,84 @@
...
+                  <li>
+                    <p>
+                      Request user consent to use <var>accumulated configuration</var> in the <var>origin</var> and wait for the user response.
+                    </p>
+                    <p class="note">User consent to use<var>accumulated configuration</var> is dependent at least on the value of the <a def-id="option-distinctiveIdentifier"></a> member.</p>
+                    <p class="note">User consent to use <var>accumulated configuration</var> is specific to the <var>origin</var> and may be limited to configurations sharing certain properties with <var>accumulated configuration</var>.</p>
+                  </li>
+                  <li>
+                    <p>
+                      If consent was denied, add the specific configurations for which consent was denied to <var>restrictions</var>
```
ddorwin: Same comment as line 657 about "configurations."

encrypted-media-respec.html
ddorwin: See other references to "configurations" and this line in my comments below.

I think it's possible to deny a configuration without denying everything in that configuration in future iterations. For example, if an audio codec happens to be in an _accumulated configuration_ that is denied because "optional" was converted to "required", that doesn't mean that the audio codec should be disallowed in future iterations.

ddorwin:
Thanks. This looks good overall. There are some details to work out and nits in the comments.

FYI, I think it will be quicker to iterate on this if the minor changes are in a single commit and rebasing is delayed until the end.

encrypted-media-respec.html
mwatson2: I suggest doing two things:
- replace the "Continue" step above for both DI and PS with a step which downgrades the parameter based on _restrictions_
- Move the first part of this para to an explicit step above step 2. Existing step 2 then becomes "if _implementation_ does not support _session type_ ... *for other reasons* ..."

encrypted-media-respec.html
mwatson2: I believe correct syntax for a list is "a, b, c and d", so if we have used "a, b, c*,* and d" elsewhere, it's that one which is wrong.

encrypted-media-respec.html
mwatson2: This one is odd, because it is a requirement that is relevant only for non-compliant implementations and so one wonders why non-compliant implementations would take the trouble to be compliant to this.

Anyway, _restrictions_ records all the values of _accumulated configuration_ for which the user has denied consent (or, at least, it is a rule which can determine whether a given specific value of _accumulated configuration_ has been denied consent - see above). In this case, the specification is effectively denying consent on behalf of the user, so  it makes sense to update _restrictions_. If we don't won't we just come straight back around to here again ?

When we add _accumulated configuration_ to restrictions we add that single combination of parameters, codecs, robustness etc. We may well come around again with one that is only very slightly different, but eventually we'll run out of options that trigger this check and take a different path.

encrypted-media-respec.html
ddorwin: SGTM.

encrypted-media-respec.html
ddorwin: Correct/wrong depends on whom you ask: https://en.wikipedia.org/wiki/Serial_comma. :)

I always use the serial comma, so it is probably used more often than not in the spec.

encrypted-media-respec.html
mwatson2: Removed "an". I'll deal with Oxford comma consistency in a separate commit.

encrypted-media-respec.html
mwatson2: See my response to that comment.

mwatson2:
Ok, I did all the changes based on the last set of review comments in one commit and rebased. Are we good to go ?

encrypted-media-respec.html
```diff
@@ -608,11 +608,11 @@
...
               <p>
               Supported configurations, including supported audio and video codecs, may depend on availability of optional capabilities such as
-              <a def-id="distinctive-identifier"></a>s and persistent state. The following algorithm iteratively tries to find a configuration
-              that is both supported and has user consent, each iteration considering a stronger set of restrictions derived from user consent prompts.
+              <a def-id="distinctive-identifier">Distinctive Identifier(s)</a> and persistent state. The following algorithm iteratively tries to find a configuration
+              that is both supported and has user consent, each iteration considering a smaller set of possible configurations.
               </p>
               <p>
-              User Agents should persist consent responses at least for the duration of the <a def-id="requestMediaKeySystemAccess"></a>
+              User Agents should reuse earlier consent responses at least for the duration of the <a def-id="requestMediaKeySystemAccess"></a>
```
ddorwin: I think we should "when/where appropriate" after "responses", acknowledging that there are some nuances (discussed later).

encrypted-media-respec.html
```diff
@@ -694,15 +694,23 @@
                   </dd>
                   <dt><a def-id="requirement-optional"></a></dt>
                   <dd>
-                    <p>Continue.<!-- Will be updated below. --></p>
+                    <p>If <a def-id="distinctive-identifiers"></a> are not allowed according to <var>restrictions</var>:</p>
```
ddorwin: nit: For consistency, add ", run the following steps:" Same at line 723.

encrypted-media-respec.html
```diff
@@ -694,15 +694,23 @@
...
+                    <ol>
+                      <li>
+                        <p>Set the <a def-id="option-distinctiveIdentifier"></a> member of <var>accumulated configuration</var> to <a def-id="requirement-not-allowed"></a></p>
```
ddorwin: missing period at end

encrypted-media-respec.html
```diff
@@ -694,15 +694,23 @@
...
+                      </li>
+                      <li>
+                        <p>If the <var>implementation</var> requires a <a def-id="distinctive-identifier"></a> in combination with <var>accumulated configuration</var> and <var>restrictions</var>, return <code>NotSupported</code>.</p>
```
ddorwin: This is really executing the "not-allowed" case. I wonder if it might be clearer and less susceptible to rot if we said, "Repeat this step" or "Execute the steps for the `<a def-id="requirement-not-allowed"></a></dt>` case." Thoughts?

encrypted-media-respec.html
```diff
@@ -694,15 +694,23 @@
...
+                    </ol>
                   </dd>
                   <dt><a def-id="requirement-not-allowed"></a></dt>
                   <dd>
                     <p>If the <var>implementation</var> requires a <a def-id="distinctive-identifier"></a> in combination with <var>accumulated configuration</var> and <var>restrictions</var>, return <code>NotSupported</code>.</p>
                   </dd>
                 </dl>
               </li>
-              <li><p>Set the <a def-id="option-distinctiveIdentifier"></a> member of <var>accumulated configuration</var> to equal <var>candidate configuration</var>'s <a def-id="option-distinctiveIdentifier"></a> member.
+              <li><p>If the <a def-id="option-distinctiveIdentifier"></a> member of <var>accumulated configuration</var> has not been set, set it to equal <var>candidate configuration</var>'s <a def-id="option-distinctiveIdentifier"></a> member.
```
ddorwin: Alternatively, we could insert a step to set a local variable to the "`the value of <var>candidate configuration</var>'s <a def-id="option-distinctiveIdentifier"></a> member.`" then just set accumulated configuration's value to that.

Although it adds a step, I think it's simpler than setting accumulated configuration's value in two places.

encrypted-media-respec.html
ddorwin: Actually, that doesn't work because we are modifying a different value than was checked. One solution is to use a local variable per my comment at line 713.

Alternatively, we could avoid this complexity by extracting this into a step before the "switch statement." We would need to restore the "Continue..." and check for "optional" twice, but it might be less complex.

For example:
* `Let <var>distinctive identifier requirement</var> be the value of <var>candidate configuration</var>'s <a def-id="option-distinctiveIdentifier"></a> member.`
* `If  <var>distinctive identifier requirement</var> is <a def-id="requirement-optional"></a> and <a def-id="distinctive-identifiers"></a> are not allowed according to <var>restrictions</var>, let <var>distinctive identifier requirement</var> be <a def-id="requirement-not-allowed"></a>.`
* Execute steps before this commit...

encrypted-media-respec.html
```diff
@@ -596,8 +596,56 @@
...
+              consent to one configuration may also imply consent for less powerful, more restricted configurations. Equally, a denial of consent for
+              one configuration may imply denial of consent for more powerful, less restricted configurations.
+              </p>
+              <p>
+              Supported configurations, including supported audio and video codecs, may depend on availability of optional capabilities such as
+              <a def-id="distinctive-identifier">Distinctive Identifier(s)</a> and persistent state. The following algorithm iteratively tries to find a configuration
+              that is both supported and has user consent, each iteration considering a smaller set of possible configurations.
+              </p>
+              <p>
+              User Agents should reuse earlier consent responses at least for the duration of the <a def-id="requestMediaKeySystemAccess"></a>
```
ddorwin: I think we should "when/where appropriate" after "responses", acknowledging that there are some nuances (discussed later).

encrypted-media-respec.html
```diff
@@ -741,73 +809,63 @@
...
 
-              <li><p>If <var>implementation</var> in the configuration specified by the combination of the values in <var>accumulated configuration</var> is not supported or not allowed in the <var>origin</var>, return <code>null</code>.</p>
+              <li><p>If <var>implementation</var> in the configuration specified by the combination of the values in <var>accumulated configuration</var> is not supported or not allowed in the <var>origin</var>, return <code>NotSupported</code>.</p>
                 <p class="note">In this step, "supported" includes the implementation being available for use when this algorithm returns, not just user agent support for such an implementation.</p>
               </li>
 
-              <li><p>If <var>accumulated configuration</var>'s <a def-id="option-distinctiveIdentifier"></a> value is <a def-id="requirement-required"></a>:</p>
+              <li>
+                <p>
+                  If <var>accumulated configuration</var>'s <a def-id="option-distinctiveIdentifier"></a> value is <a def-id="requirement-required"></a> and if
```
ddorwin: nit: I think we can remove the last "if" on this line.

encrypted-media-respec.html
ddorwin: This last paragraph helps me understand your intent with _restrictions_. I'll continue this discussion above (line 648).

encrypted-media-respec.html
mwatson2: UAs can ask for consent in a way that collects consent for more than just the specific _accumulated configuration_ with a single question (depends on how they choose to word the question and interpret the response). But the consent should at least be specific to the value of `distincitiveIdentifier` even if nothing else.

I think there was a pre-existing note to this effect which I may have re-worded (badly ?)

encrypted-media-respec.html
ddorwin: The last paragraph of your comment at line 824 sheds some light on your intent:
>When we add _accumulated configuration_ to restrictions we add that single combination of parameters, codecs, robustness etc.

My thought was that we would only add specific things that were denied. For example "Distinctive Identifiers are disallowed" or (much less likely) "Distinctive Identifiers are disallowed with codec foo."

I see the following issues with adding an entire denied configuration:
* The individual steps are much more complex:
 * The current text of "`in combination with <var>accumulated configuration</var> and <var>restrictions</var>`" is not accurate if _restrictions_ is a list of configurations.
* Most importantly, I'm not sure how exactly to determine whether a choice being made at an individual step is compatible.
 * The only way to make a comparison is to compare the entire configuration at the end, but that's too late and we'll loop indefinitely.
* Use of a normative structure prevents remembering implementation-specific rules.

mwatson2: Right, consent is denied to specific values of _accumulated configuration_: that audio codec with `required` DIs was denied, but the same audio codec with a different DI value might be allowed.

The UA is free to interpret the response as covering other configuration values as well. In this case, the UA might also add configurations that combine other audio codecs with the same `required` DI value. This is dependent on the question the UA chooses to ask. If the user is asked "Allow _keySystem_ access to distinctive identifiers with audio codec AXX" maybe only that specific combination is denied, but if (more likely) they are just asked "Allow _keySystem_ access to distinctive identifiers", maybe a whole slew of possible combinations are recorded as denied.

encrypted-media-respec.html
ddorwin: Bump.

encrypted-media-respec.html
mwatson2: > Most importantly, I'm not sure how exactly to determine whether a choice being made at an individual step is compatible.

It's a good question. Where we say "_implementation_ supports X in combination with _accumulated configuration_ and _restrictions_." it means there is some supported final configuration which has everything _accumulated configuration_ has as well as X and which is not in _restrictions_.

One would not implement _restrictions_ as an actual exhaustive list, but as a structure with implementation-specific semantic elements - probably along the lines you are thinking. But logically it is still a rule which tells us whether any given configuration has been denied or not and logically this is equivalent to the list of _configuration_ values.

I think any given implementation-specific rule can be handled through the flexibility to add whatever configurations the implementation thinks appropriate.

encrypted-media-respec.html
ddorwin: The relevant step previously said, "Request user consent to use Distinctive Identifier(s)..." I'm guessing you were noting that this more general consent must at least consider DI.

It might be more straightforward to say something like `"The consent MUST include consent to use a DI if <var>accumulated configuration</var>'s <a def-id="option-distinctiveIdentifier"></a> member is <a def-id="requirement-required"></a>.`"

ddorwin:
Thanks. My feedback falls into three categories:

1. What _restrictions_ is and how it is used in the individual steps.
1. Various options for applying _restrictions_ to `"optional"`.
1. Nits.

FYI, I'm trying to make all my changes in the "Files changed" tab (`/files`). This appears to maintain old comments from unchanged lines. I also ran into a problem where not all comments made to the last commit show up there for some reason and had to move the comment.


<br/>Also, to clarify my earlier statement, I meant we should rebase after the review process is complete. Related to that...

One thing I've discovered recently is that GitHub _does_ delete commits on forced pushes. The link for "View full outdated diff" ([example](https://github.com/w3c/encrypted-media/pull/165/files/1c00a60beb5887fb7cd0f97174727ce415a49058#r60295455)) even tells you this.

Also, it appears [GitHub now supports squashing via the big green button!](https://github.com/blog/2141-squash-your-commits). HTML now [allows use](https://github.com/whatwg/html/blob/master/TEAM.md) of the button. We should verify our configuration, try using the button, and update our process. That should pretty much eliminate forced pushes in most cases I think (assuming rebases are handled correctly).

mwatson2:
@ddorwin I've implemented your suggestions for (2) and (3) (not pushed yet).

For (1), I've added some comments, but I am not sure what is the best path for modifying the text.

If we wanted to be much more explicit, we could replace _restrictions_ with an algorithm "Is consent denied for _configuration_ in _origin_ ?". Then, phrases such as:
> If _implementation_ supports X in association with _accumulated configuration_ and _restrictions_ ...

would become
> If there exists a _configuration_ derived from _accumulated configuration_ that supports X, is supported by _implementation_ and for which the "Is consent denied for _configuration_ in _origin_ ?" algorithm returns `false`, then ...

(We would also define what it means for _configuration_ to be derived from _accumulated configuration_ which is just that for everything except `audioCapabilities` and `videoCapabilities` every member of _accumulated configuration_ appears in _configuration_ with the same value and for `audioCapabilities` and `videoCapabilities` every element that appears in _accumulated configuration_ also appears in _configuration_.)

We could then leave the details of "Is consent denied for _configuration_ in _origin_ ?" mostly to the implementation.

I'm not sure this extra level of detail is necessary, though. WDYT ?

mwatson2:
That's nice about the green button, but we still need to rebase, which may then necessitate a force push.

I'm not sure what the best plan is there ? We could create a new branch in which to do the rebase, generate a new PR and merge that ?

encrypted-media-respec.html
```diff
@@ -899,42 +957,84 @@
...
+                  If any of the following are true:
                 </p>
+                <ul>
+                  <li>
+                    <p>
+                      The <a def-id="option-distinctiveIdentifier"></a> member of <var>accumulated configuration</var> is not
+                      <a def-id="requirement-not-allowed"></a> and any of the recommendations of
+                      <a href="#allow-identifiers-cleared">Allow Identifiers to be Cleared</a> are not supported by the combination of the
+                      User Agent and <var>implementation</var>.
+                    </p>
```
mwatson2: I believe this should be "User agent, _implementation_, and *_accumulated configuration_*" since _accumulated configuration_ may contain `robustness` values that imply particular CDM properties (e.g. hardware vs software) which in turn may have different Distinctive Identifier properties.

ddorwin:
It seems this complexity is caused by trying to use and select an entire configuration.

I think it would be easier if we treat _restrictions_ as a collection of "implementation-specific semantic elements" (as you said in another comment, though I'm not proposing we use this language in the spec) and make decisions with those in mind ("supports X in association with... _restrictions_"). How the user agent applies those or makes decisions is implementation-specific as long as selecting X would not cause the configuration to be covered by the denied consent (as represented in _restrictions_).

We could also consider replacing "and _restrictions_" with something like "and the resulting configuration would not be denied based on the _restrictions_."

Is there a reason you prefer the configurations model? Given that it is harder to comprehend (IMO), especially when thinking in terms of implementation, and doesn't handle implementation-specific rules, I think the collection of rules model is a better option.


<br/>I'll move the green button discussion to email where it will be easier to track.

mwatson2:
@ddorwin I added some notes in my last commit which move closer to the language you suggest. I still think our approaches are equivalent: _restrictions_ is a rule which can tell you if a given configuration is denied and such a rule is exactly equivalent to the set of specific configurations that are denied (in particular, the proposed text _does_ support implementation specific rules).

Nevertheless, given that the approaches are equivalent, I'm happy to move over to "rule-oriented" rather than a "set-oriented" language. The latter seemed more concrete to me, but the "rule" approach can be made so as well.

mwatson2:
I've relaxed the language around _restrictions_ to say that it represents the configurations for which consent has been denied without saying how it does that. Rather than "add _configuration_ to _restrictions_" we now say "update _restrictions_ to reflect the configurations for which consent has been denied"

encrypted-media-respec.html
```diff
@@ -609,19 +609,24 @@
...
-              that is both supported and has user consent, each iteration considering a smaller set of possible configurations.
+              that is both supported and has user consent (or does not need consent).
               </p>
               <p>
               User Agents should reuse earlier consent responses, when appropriate, at least for the duration of the <a def-id="requestMediaKeySystemAccess"></a>
               algorithm in order to avoid repeated requests to the user for similar configurations.
               </p>
+              <p>
+                The variable <var>restrictions</var> in the steps below represents the configurations for which consent has been denied. It is used to determine
+                whether user consent for a candidate configuration or partial configuration has been denied. Consent is denied for a partial configuration if every derived
```
ddorwin: Perhaps s/has been denied/would be denied based on previous {information obtained in previous iterations}/

The text in {} may need improvement, but the important part is past=>future for _this_ configuration.

mwatson2: I'm not sure about this. "... would be denied ..." implies a prediction about what the user _would_ say if we asked them, and we cannot know that (they might change their mind). We've explained above that a single consent prompt can be interpreted as denying a whole slew of configurations based on exactly what the user was asked and how they responded so I think it's reasonable to say _... consent has been denied ..._.

Would it help to say _... consent has been denied during the execution of this algorithm or based on persisted consent information for the origin._ ?

mwatson2: I added this clarification. Are we ready to merge this one ?

encrypted-media-respec.html
```diff
@@ -638,42 +691,83 @@
...
                   </dd>
                   <dt><a def-id="requirement-optional"></a></dt>
                   <dd>
-                    <p>Continue.<!-- Will be updated below. --></p>
+                    <p>Continue with the following steps.</p>
                   </dd>
                   <dt><a def-id="requirement-not-allowed"></a></dt>
                   <dd>
-                    <p>If the <var>implementation</var> requires a <a def-id="distinctive-identifier"></a> in combination with <var>accumulated configuration</var>, return <code>null</code>.</p>
+                    <p>If the <var>implementation</var> requires a <a def-id="distinctive-identifier"></a> in combination with <var>accumulated configuration</var> and <var>restrictions</var>, return <code>NotSupported</code>.</p>
```
ddorwin: Related to the next comment:
Looking at the (existing and new) text more closely, it may not be clear/correct. "A requires X in combination with B and C." I believe it is A, B, and C that are the combination and X is the result. Or A requires X when B and C. Either way, "in combination with" seems kludgy.
One option would be to worded these as "If X is required by A when in the configuration B with restrictions C applied." It's a bit wordy, though. However, this _might_ eliminate the need for the note below.

This should probably be addressed separately since it already exists and is unrelated to these changes, though they exacerbate the issue.

encrypted-media-respec.html
```diff
@@ -638,42 +691,83 @@
...
                   </dd>
                 </dl>
+                <div class="note">
+                <p>
+                  The combination of <var>accumulated configuration</var> and <var>restrictions</var> means all the possible configurations that
```
ddorwin: I understand what you are trying to communicate here, but "combination" is used differently here than in the steps. (See above for my concerns about the use of that word.) I don't believe there is a combination of these two values (B and C above) in isolation. It's at least weird to use "combination" in two different ways.

Should we just add a TODO to fix this along with the issues in the previous comment? Can we eliminate the use of "combination" here?

encrypted-media-respec.html
```diff
@@ -638,42 +691,83 @@
...
+                  include everything in <var>accumulated configuration</var> and that are not denied according to <var>restrictions</var>.
+                </p>
+                <p>
+                  A feature is supported with this combination if at least one of the configurations in the combination supports the feature.
```
ddorwin: The "combination" referenced here doesn't exist. It's B and C but we also need A. I think this is really referring to the set of all possible configurations that satisfy _accumulated configuration_ after applying _restrictions_.

encrypted-media-respec.html
```diff
@@ -596,8 +596,61 @@
...
+                whether user consent for a candidate configuration or partial configuration has been denied. Consent is denied for a partial configuration
+                if every derived configuration has already been denied. Internal representation of <var>restrictions</var> is implementation-specific.
+              </p>
+            </div>
+            <ol>
+              <li>
+                <p>Let <var>supported configuration</var> be <code>ConsentDenied</code>.</p>
+              </li>
+              <li>
+                <p>Initialize <var>restrictions</var> to represent existing consent information for <var>origin</var> if any.</p>
```
ddorwin: Note: This conceptually changes where persisted consent is checked. I'm not sure whether it has an observable distinction.

encrypted-media-respec.html
```diff
@@ -899,46 +992,94 @@
...
+                Previous consent for <var>accumulated configuration</var> with <a def-id="option-distinctiveIdentifier"></a>
+                set to <a def-id="requirement-not-allowed"></a> does not imply consent for the same configuration with <a def-id="option-distinctiveIdentifier"></a>
+                set to <a def-id="requirement-optional"></a> or <a def-id="requirement-required"></a>.
+              </p>
+            </div>
             <ol>
               <li>
                 <p>
-                  If there is persisted denial for <var>origin</var> indicating that <a def-id="distinctive-identifier">Distinctive Identifier(s)</a> are not allowed in conjunction with <var>accumulated configuration</var>, return <code>Denied</code>.
+                  If there is persisted denial for <var>origin</var> indicating that <var>accumulated configuration</var> is not allowed:
```
ddorwin: Is this step necessary with the change in line 630?
> <p>Initialize <var>restrictions</var> to represent existing consent information for <var>origin</var> if any.</p>

ddorwin: nit:
>, run the following steps:

encrypted-media-respec.html
```diff
@@ -899,46 +992,94 @@
...
+                      Request user consent to use <var>accumulated configuration</var> in the <var>origin</var> and wait for the user response.
+                    </p>
+                    <div class="note">
+                      <p>The consent must include consent to use a <a def-id="distinctive-identifier"></a> if <var>accumulated configuration</var>'s <a def-id="option-distinctiveIdentifier"></a> member is <a def-id="requirement-required"></a>.</p>
+                      <p>User consent to use <var>accumulated configuration</var> is specific to the <var>origin</var> and may be limited to configurations sharing certain properties with <var>accumulated configuration</var>.</p>
+                    </div>
+                  </li>
+                  <li>
+                    <p>
+                      If consent was denied, update <var>restrictions</var> to reflect the configurations for which consent was denied
```
ddorwin: nit: On line 1015, the update and return are separate steps. We should be consistent.

encrypted-media-respec.html
mwatson2: I'll revert it. We can change it later if we think we need to.

encrypted-media-respec.html
```diff
@@ -741,73 +839,68 @@
...
+                  <dd>
+                    <p>
+                      Continue to the next step.
+                    </p>
+                  </dd>
+                </dl>
               </li>
 
-              <li><p>Return <var>accumulated configuration</var>.</p></li>
+              <li><p>Return <var>accumulated configuration</var></p></li>
```
ddorwin: nit: The period was removed.

encrypted-media-respec.html
mwatson2: I think what we mean is 

for _requires_: "If A requires X in all configurations derived from B for which the user has not denied consent according to C"

for _supports_: "If A supports a configuration derived from B that supports X and for which the user has not denied consent according to C."

However, I don't think the combination word is so bad. It is possible to "combine" two things of different types to produce a result of a third type, so long as the operator ("combine") is defined for that pair of input types.

So, I suggest leaving a TODO here for the moment.

encrypted-media-respec.html
```diff
@@ -899,46 +992,94 @@
...
+                  </li>
+                </ul>
+                <p>then run the following steps:</p>
+                <ol>
+                  <li>
+                    <p>
+                      Request user consent to use <var>accumulated configuration</var> in the <var>origin</var> and wait for the user response.
+                    </p>
+                    <div class="note">
+                      <p>The consent must include consent to use a <a def-id="distinctive-identifier"></a> if <var>accumulated configuration</var>'s <a def-id="option-distinctiveIdentifier"></a> member is <a def-id="requirement-required"></a>.</p>
```
ddorwin: Should this be part of the normative text with a normative MUST?

encrypted-media-respec.html
mwatson2: Yes. Need to include _implementation_ in those definitions.

ddorwin:
I think the first three comments above (since your last commit) can be handled as TODOs so we can land this PR. Then there is a change we agree on and some nits. Other than that, LGTM. Thanks for tackling this complex change!

mwatson2:
Merged as https://github.com/w3c/encrypted-media/commit/ff17c6b37ea8c0df475e2bbd133fc78229c836ee

Issue #105, opened 2015-10-19
EME registry: Separate definitions of Initialization Data Types from Stream Format parsing


ddorwin:
Originally filed as [bug 26811](https://www.w3.org/Bugs/Public/show_bug.cgi?id=26811).

In summary, the EME registry was created before Initialization Data Types were created. Now, the MIME type and Initialization Data Type are different and may be independent. Thus, we should have separate registries.

I believe we should also put the registries in their own directories. This will make publishing the registry separate from the EME spec (issue #104) easier and shorten all the filenames. For example:
```encrypted-media\
  ...
  registry\
    stream-format\
      bmff.html
      webm.thml
      ...
    initdata-format\
      cenc.html
      webm.thml
      ...
```

ddorwin:
@plehegar FYI related to how this will be published.

paulbrucecotton:
@plehegar - Any update on your work on this item?
See the Sapporo minutes where you agreed to review this:
http://www.w3.org/2015/10/30-html-media-minutes.html#item26

plehegar:
@paulbrucecotton I should get around this within 2 weeks

ddorwin:
This issues is really on me to break these up.

Any comments from the community on the proposed structure?

boblund:
I proposed a structure here https://www.w3.org/Bugs/Public/show_bug.cgi?id=26811#c3 for a cenc init doc and a BMFF encapsulation doc. Another proposal https://www.w3.org/Bugs/Public/show_bug.cgi?id=26738#c21 uses the cenc init doc with a proposed MPEG2TS encapsulation doc.

ddorwin:
Thanks. I had in mind a more complete separation where each type of doc has different sections. I can create a PR, but it would be helpful to have feedback on the actual file structure before I start.

Separately, `bmff.html` might need to be `mp4.html` given the link to mime types.

jdsmith3000:
I thought I commented here earlier, but my comment is missing.  Roughly restating it:  The ISOBMFF and CENC combination is pretty tightly coupled, and separating the specs doesn't seem useful.  Can you provide other cases where this separation is needed?

ddorwin:
[...]

PR #167, opened 2016-04-15
Issue #105: Separate registry definitions of Initialization Data Types and stream formats

ddorwin:
The first commit copies/moves the files to the new locations. There are no changes. Unfortunately, GitHub does not appear to track this and shows that all files in `stream/` are entirely new when viewing files changed.

The ReSpec output files (those that don't end in `-respec.html`) have not been updated. I will do that before landing.

If this looks good overall, I propose that we land it and continue to iterate as necessary. Comments welcome, of course, and I'll try to address them in this PR when it makes sense.


ddorwin:
I fixed up the last two commits so they appear in the correct order in GitHub. This includes a minor text change to the index pages.

ddorwin:
Merged as the four commits prior to e8e3284, which updated the generated files.

Issue #171, opened 2016-04-25
Clarify steps for destruction of keys / licenses


mwatson2:
Discussion on #85, (see, in particular, [this comment](https://github.com/w3c/encrypted-media/issues/85#issuecomment-211496666)), suggests the specification could describe the steps for destruction of keys / licenses more clearly.

I suggest we:
(1) extract the steps for destruction of keys / licenses from the _Session close_ algorithm into a new _Destroy keys / licenses_ algorithm.
(2) refer to this algorithm from the `persistent-usage-record` session type definition (instead of the current text which duplicates the requirements in prose).

These changes would not add or remove any requirements, but would make the existing requirements easier to understand.

paulbrucecotton:
@ddorwin, @jdsmith3000:  Can one of you please review and merge in Mark's proposed changes so that we can close this issue?

ddorwin:
For those not following the discussion in the PR, I don't think we should include key destruction in a user agent algorithm invoked by the CDM. See [this comment](https://github.com/w3c/encrypted-media/pull/172/files#r61828575)).

mwatson2:
I will look at this more closely tomorrow, but it was not my intention that
the algorithm be invoked by the CDM, but rather that the User Agent invokes
it when the MediaKeySession is destroyed and tells the CDM (asynchronously
if the user agent implementation wishes) to destroy the keys.

As an aside, I don't think our specification can, does or should address
the relationship between UA and CDM in terms of execution model (I.e.
threading). We define only the observable behavior and anything else is
implementation-specific. Typically we say that the user agent 'uses the
CDM' (usually asynchronously to the JS thread) to perform some task, but we
do not say what that involves, implementation-wise.

As an example, if some task or event is triggered by the CDM (for example a
license renewal message), we do not say whether this is implemented as a
callback from the CDM, the CDM posting a task to some kind of queue, the UA
polling the CDM or any other implementation approach.

On May 2, 2016, at 6:23 PM, ddorwin <notifications@github.com> wrote:

For those not following the discussion in the PR, I don't think we should
include key destruction in a user agent algorithm invoked by the CDM. See this
comment <https://github.com/w3c/encrypted-media/pull/172/files#r61828575>).

—
You are receiving this because you authored the thread.
Reply to this email directly or view it on GitHub
<https://github.com/w3c/encrypted-media/issues/171#issuecomment-216411073>

mwatson2:
[...]

Issue #153, opened 2016-03-25
Non-method algorithm intro paragraphs may be misleading about when to run the algorithm


ddorwin:
(The following focuses on added HTMLMediaElement algorithms, but this also applies to algorithms for other objects.)

The section for several algorithms in the spec begin with text like "The following steps are run when..." "The Do Something algorithm is run when..."

For some algorithms, such as the [Initialization Data Encountered](https://w3c.github.io/encrypted-media/#initdata-encountered) algorithm, this means the algorithm should be run when this happens. However, for others, the when to call the algorithms is explicitly specified somewhere else and this is just informative context. There is no differentiation of these cases.

First, we should avoid implying the former when it is not the intent (the latter). Instead, we can follow other examples, such as "This algorithm does something..." Second, we _may_ not even need the former. For example, some recent updates explicitly added when to call the "Encrypted Block Encountered" algorithm to the top of https://w3c.github.io/encrypted-media/#htmlmediaelement-extensions.

mwatson2:
This is expected to be solved by #181

mwatson2:
Closing because we'll address this under #181.

ddorwin:
The "run when" text still appears in three HTMLMediaElement algorithms:
* Initialization Data Encountered
* Queue a "waitingforkey" Event
* Attempt to Resume Playback If Necessary

I think we addressed the `MediaKeySession` algorithms in #181, but these were not addressed.

ddorwin:
PR #285.

Issue #85, opened 2015-08-27
"tracked" sessions: architectural concerns pending resolution with TAG


ddorwin:
Pull request #54 was merged [without](http://www.w3.org/2015/08/18-html-media-minutes.html#item01) [addressing](https://lists.w3.org/Archives/Public/public-html-media/2015Aug/0025.html) architectural concerns about “tracked” sessions. Unresolved questions are pending a discussion with the TAG. The outcome could result in modification (or removal) of “tracked” sessions. 

This issue is a placeholder for that discussion and outcome.

Resolving #82 and #84 could help accelerate conclusion of this discussion.

mwatson2:
The group did not agree that the issue needed to be raised by the TAG. Of course companies are free to do so and that might result in TAG advice for the group to further consider. But it is not the case that we are blocked pending discussion with the TAG.

ddorwin:
@mwatson2: This issue was carefully worded to reflect the situation. While I am unable to find a path to your conclusion that it implies that "tracked" is "blocked pending discussion with the TAG," I want to clarify that you are right - "tracked" isn't blocked pending a TAG discussion. However, I also believe that this discussion '*could* result in modification (or removal) of “tracked” sessions.' Do you disagree?

mwatson2:
My comment was just to clarify that the unresolved questions and concerns are Google's, not shared by the group, as this wasn't stated either way in the issue description.

mwatson2:
I offered on our call to provide information about the fraction of time we expect to / do receive the secure release information. I have a qualified answer: considering only streaming sessions where either:
(a) the session is closed gracefully and the secure release exchange with the server completes, or
(b) (a) does not hold, but the user later revisits the site and Local Storage information from the session has not been cleared
we expect to receive secure release messages for at least 99% of sessions. And indeed in practice we achieve this in the field a desktop browser that has implemented secure release.

paulbrucecotton:
As the Sapporo F2F meeting we have no update from the TAG in either EME ISSUE-85 or the related TAG Issue-73.

Paul will continue to chase after @travisleithead and Alex to get feedback on this matter.

travisleithead:
@slightlyoff and I are still discussing this. I hope we can make some progress this week.

paulbrucecotton:
@travisleithead and @slightlyoff:  Can you give us an update on your progress?

paulbrucecotton:
[...]

PR #172, opened 2016-04-25
Issue #171: Clarify destruction of licenses and keys

mwatson2:


jdsmith3000:
This change looks good to me.  It creates a new **Destroy Keys and Licenses** algorithm that is substantially equivalent to the current spec behavior, with two notable (and to me desirable) clarifications:

1. It adds that the algorithm MUST be executed when the `MediaKeySession` object becomes inaccessible to the page, if **Session Close** has not already been run.  Running **Session Close** was the sole trigger for these steps previously.
2. It adds a note that keys and licenses may be destroyed asynchronously, including after the document has been unloaded.

@ddorwin I believe this PR can be merged and will do so later today, assuming no objections from you.

@mwatson Minor correction:  Text currently says "if the Session Close has not already run".  Should remove "the".

mwatson2:
@jdsmith3000 
> Running Session Close was the sole trigger for these steps previously.

Actually, previously the destruction of keys and licenses was explicitly required for `temporary-usage-record` sessions by the test definition of that enum value. It was not previously required for `temporary` sessions.

> Minor correction: Text currently says "if the Session Close has not already run". Should remove "the".

Should be "...the Session Close algorithm..."

encrypted-media-respec.html
```diff
@@ -1158,7 +1158,7 @@
         <dt>temporary</dt>
         <dd>
           <p>
-            A session for which the license and record of or data related to the session MUST NOT be persisted.
+            A session for which the license, key(s) and record of or data related to the session are not persisted.
```
ddorwin: nit: serial comma debate :)

encrypted-media-respec.html
```diff
@@ -1169,7 +1169,7 @@
         <dd>
           <p class="issue"><a href="https://github.com/w3c/encrypted-media/issues/85">Issue 85</a> - There are concerns about the architectural implications of <code>"persistent-usage-record"</code> sessions that are pending a TAG discussion. The outcome of this discussion could result in modification (including removal) of the feature.</p>
           <p>
-            A session for which the license and any key(s) it contains SHALL NOT be persisted and for which a <a def-id="record-of-key-usage"></a> SHALL be persisted when
+            A session for which the license and key(s) are not persisted and for which a <a def-id="record-of-key-usage"></a> is persisted when
```
ddorwin: Is there a reason you switched to non-RFC words? I suppose that reads better as a description, though we do want these to be normatively required behaviors (hopefully also defined in the algorithms). I wonder if these descriptions are supposed to be normative, though.

There are still some RFC words in the following text. In particular, the "SHALL be" for `persistent-license` seems equivalent to the "SHALL BE" that you replaced here. These should be consistent either way.

mwatson2: Yes, the normative requirement is moved to the _Destroy keys and licenses_ algorithm (Section 6.6.5) and so this text becomes descriptive so as not to have duplicate specification of the same normative requirement.

And yes, it would make sense to move the requirement to store the record of license destruction for `persistent` sessions as well.

encrypted-media-respec.html
```diff
@@ -2101,6 +2101,30 @@
...
             <li><p>Let the <var>session</var> be the associated <a>MediaKeySession</a> object.</p></li>
+            <li><p>Run the <a def-id="destroy-keys-and-licenses-algorithm"></a> algorithm.</p></li>
+            <li><p>Run the <a def-id="update-key-statuses-algorithm"></a> algorithm on the <var>session</var>, providing an empty sequence.</p></li>
+            <li><p>Run the <a def-id="update-expiration-algorithm"></a> algorithm on the <var>session</var>, providing <code>NaN</code>.</p></li>
+            <li><p>Let <var>promise</var> be the <a def-id="closed"></a> attribute of the <var>session</var>.</p></li>
+            <li><p>Resolve <var>promise</var>.</p></li>
+          </ol>
+        </section>
+        <section id="destroy-keys">
+          <h4>Destroy Keys and Licenses</h4>
```
ddorwin: Since this only applies to non-persistent/in-memory keys and licenses, should the name be explicit about this?

encrypted-media-respec.html
```diff
@@ -2101,6 +2101,30 @@
...
+          <p>
+            The Destroy Keys and Licenses algorithm is run when the CDM needs to destroy all the keys and licenses associated with a session that
```
ddorwin: As written, there seems to be overlap between when these two algorithms are run.

Also, we may want to minimize the use of "... is run when..." language. In addition to the issues in #153, it creates the opportunity for ambiguity as in this case.

The second paragraph adds an additional requirement that describes when we might want this one to be called specifically, but that does not address this paragraph.

See also the comment below.

encrypted-media-respec.html
```diff
@@ -2101,6 +2101,30 @@
...
+            For sessions of type <a def-id="temporary-session"></a> and <a def-id="persistent-usage-record-session"></a>, this algorithm MUST be
+            executed when the <a>MediaKeySession</a> object becomes inaccessible to the page, if the <a def-id="session-close-algorithm"></a>
+            algorithm has not already been run.
+          </p>
+          <p class="note">
+            This is necessary to ensure that the keys and licenses associated with the session are correctly destroyed when the session ends.
+          </p>
+          <p>The following steps are run:</p>
+          <ol>
+            <li><p>Let the <var>session</var> be the associated <a>MediaKeySession</a> object.</p></li>
```
ddorwin: Regarding line 2135 (applies to the existing spec text too):

This algorithm, like Session Close, is an asynchronous callback from the CDM to the user agent to initiate application-visible behavior. It doesn't make sense for the CDM to call back into the user agent only to have it call back to destroy the keys. Destroying the keys is something that should happen around this time, but a) the user agent can't destroy them, which is what the step currently says, and b) this should have been done before calling the user agent.

All these algorithms have "is run when the CDM..." language, though I don't know if that is the only time they are used (see #153). This is the only one that calls back into the CDM.

One solution is to state that the CDM must have destroyed the (non-persistent) key(s) and license(s) before calling either of these two algorithms. That probably changes the name and purpose of this algorithm.

encrypted-media-respec.html
```diff
@@ -2101,6 +2101,30 @@
...
+            The Destroy Keys and Licenses algorithm is run when the CDM needs to destroy all the keys and licenses associated with a session that
+            have not been explicitly stored.
+          </p>
+          <p>
+            For sessions of type <a def-id="temporary-session"></a> and <a def-id="persistent-usage-record-session"></a>, this algorithm MUST be
+            executed when the <a>MediaKeySession</a> object becomes inaccessible to the page, if the <a def-id="session-close-algorithm"></a>
+            algorithm has not already been run.
+          </p>
+          <p class="note">
+            This is necessary to ensure that the keys and licenses associated with the session are correctly destroyed when the session ends.
```
ddorwin: This seems like a CDM robustness and implementation detail. It should not be necessary to call this callback into the user agent in order to ensure the keys are destroyed. See also the comment at line 2127 below.

encrypted-media-respec.html
```diff
@@ -2101,6 +2101,30 @@
...
+        </section>
+        <section id="destroy-keys">
+          <h4>Destroy Keys and Licenses</h4>
+          <p>
+            The Destroy Keys and Licenses algorithm is run when the CDM needs to destroy all the keys and licenses associated with a session that
+            have not been explicitly stored.
+          </p>
+          <p>
+            For sessions of type <a def-id="temporary-session"></a> and <a def-id="persistent-usage-record-session"></a>, this algorithm MUST be
+            executed when the <a>MediaKeySession</a> object becomes inaccessible to the page, if the <a def-id="session-close-algorithm"></a>
```
ddorwin: I'm not sure it makes sense to specify behavior for an object after it has become "inaccessible to the page." Note that this is a user agent algorithm (see the comment at line 2127 below) for an object that has become "inaccessible to the page." As far as the spec is concerned, the object no longer exists and cannot be executing algorithms.

Additionally, `persistent-usage-record` requires that this algorithm be run even after the page is destroyed, which is not addressed by this text and is not currently addressable by a user agent algorithm (per the discussion #85).

encrypted-media-respec.html
```diff
@@ -2115,13 +2139,12 @@
                   <p>
                     If this object's <var>session type</var> is <a def-id="persistent-usage-record-session"></a>, store the <var>record of key usage</var>.
                   </p>
+                  <p class="note">
+                    This step may be run asynchronously, including after the document has unloaded.
```
ddorwin: In order for `persistent-usage-record` sessions to work, this note would need to apply to the entire algorithm, which is problematic as I noted at line 2119 above.

ddorwin: We probably don't want any persistence operations to block the main thread. Thus, it might make sense to replace the first part of this note by insering "run the following steps in parallel:" after the comma at line 2140.

ddorwin:
@jdsmith3000 wrote:
> This change looks good to me.  It creates a new **Destroy Keys and Licenses** algorithm that is substantially equivalent to the current spec behavior, with two notable (and to me desirable) clarifications:

>1. It adds that the algorithm MUST be executed when the `MediaKeySession` object becomes inaccessible to the page, if **Session Close** has not already been run.  Running **Session Close** was the sole trigger for these steps previously.

As I noted in my inline review comments:
* This introduces a new behavior of performing operations on an object after it has become "inaccessible to the page." There was no such (explicit) expectation for the Session Close algorithm.
* I don't believe we should be destroying keys from a user agent algorithm invoked by the user agent. Key destruction is important, but we should find another way to specify this.

>2 . It adds a note that keys and licenses may be destroyed asynchronously, including after the document has been unloaded.

Slight correction: The note was actually added to a single step for `persistent-usage-record` sessions.
As I noted in my review comments, it probably should apply to the entire algorithm like you said here. However, that is problematic for the same reasons as above.

encrypted-media-respec.html
mwatson2: Either we should change the name, or make it apply to the persistent case too per the above comment.

encrypted-media-respec.html
mwatson2: This seems to be the main issue, so I'll put all my comments here.

There are several examples where we are not all that clear about exactly how the UA and CDM interact to achieve the steps of our algorithms. The "is run when" phrase is a red flag for this issue, as you say. For this algorithm and indeed Session Close there are likely two distinct cases in any given implementation:
(a) The UA initiates the action, explicitly telling the CDM to either close a session or destroy keys associated with the session (e.g. `close()`, `MediaKeySession` destroyed, `remove()` if we include `persistent` sessions)
(b) The CDM initiates the action, informing the UA that it has either closed a session or destroyed keys associated with the session (resources lost by the CDM, other CDM-specific reasons).

I think the confusion here arises from trying to cover both cases in one piece of text.

For "Session Close", we could rename it to "Session Closed" and describe explicitly that these are the steps the UA performs after the session is closed and that they are triggered either explicitly in other algorithms of after an indication from the CDM that it has initiated closure of the session.

For "Destroy Keys.." we would split this into one part which is triggered by the UA from `remove()` (for `persistent` session) and when `MediaKeySession` is destroyed for `temporary` and `persistent-usage-record` and a second part ('Keys Destroyed') which is triggered when the CDM initiates destruction of the keys itself.

mwatson2: This seems to be the main issue, so I'll put all my comments here.

There are several examples where we are not all that clear about exactly how the UA and CDM interact to achieve the steps of our algorithms. The "is run when" phrase is a red flag for this issue, as you say. For this algorithm and indeed Session Close there are likely two distinct cases in any given implementation:
(a) The UA initiates the action, explicitly telling the CDM to either close a session or destroy keys associated with the session (e.g. `close()`, `MediaKeySession` destroyed, `remove()` if we include `persistent` sessions)
(b) The CDM initiates the action, informing the UA that it has either closed a session or destroyed keys associated with the session (resources lost by the CDM, other CDM-specific reasons).

I think the confusion here arises from trying to cover both cases in one piece of text.

For "Session Close", we could rename it to "Session Closed" and describe explicitly that these are the steps the UA performs after the session is closed and that they are triggered either explicitly in other algorithms of after an indication from the CDM that it has completed a CDM-initiated closure of the session.

For "Destroy Keys.." we would split this into one part which is triggered by the UA from `remove()` (for `persistent` session) and when `MediaKeySession` is destroyed for `temporary` and `persistent-usage-record` and a second part ('Keys Destroyed') which is triggered when the CDM completes a CDM-initiated destruction of the keys. 

In both the CDM-initiated cases, it is possible that the page happens to be closing and therefore the steps do not get executed, but in the UA-initiated case it should always be possible.

encrypted-media-respec.html
mwatson2: The User Agent needs to run these steps "_when_ the `MediaKeySession` object becomes inaccessible to the page", not _after_. `MediaKeySession` is a platform object visible to Javascript. Becoming inaccessible to the page does not mean the UA has deleted all knowledge of the associated CDM session or other state. Or, you can say that whatever causes it to become inaccessible should also cause this algorithm to be run. The following steps do not involve interaction with the page, so the fact the object is no longer JS-visible is not an issue.

encrypted-media-respec.html
mwatson2: We need to normatively specify that steps are asynchronously only when if affects observable behavior. In a method definition, we need to specify which steps occur before and which occur after the method returns. Outside a method invocation, I don't think there is any implication that steps are blocking the main JS thread (if there's no method invokation, then _where_ is there thread blocked in the Javascript code ?).

So, here we are really just noting the fact that there are no observable consequences if these steps are executed later and thus implementations are free to execute them later, similar to `<a ping>` and beacon.

encrypted-media-respec.html
```diff
@@ -2301,6 +2308,29 @@
             <li><p>Resolve <var>promise</var>.</p></li>
           </ol>
         </section>
+        
+        <section id="media-key-session-destroyed">
+          <h4>MediaKeySession destroyed</h4>
+          <p>
+            This algorithm performs steps that are necessary when a <a>MediaKeySession</a> is destroyed without
```
ddorwin: "without" somewhat implies something wrong happened. I don't think this is unexpected. I propose we simplify this a little:

>`... when a <a>MediaKeySession</a> is destroyed **and** the <a def-id="session-closed-algorithm"></a> algorithm **has not** been run.`

encrypted-media-respec.html
```diff
@@ -2301,6 +2308,29 @@
...
+            the <a def-id="session-closed-algorithm"></a> algorithm having been run.
+          </p>
+          <p>The following steps are run:</p>
+          <ol>
+            <li><p>Let the <var>session</var> be the associated <a>MediaKeySession</a> object.</p></li>
+            <li><p>Let <var>cdm</var> be the CDM instance represented by this object's <var>cdm instance</var> value.</p></li>
```
ddorwin: `s/this object's/<var>session</var>'s/`

encrypted-media-respec.html
```diff
@@ -2301,6 +2308,29 @@
...
           </ol>
         </section>
+        
+        <section id="media-key-session-destroyed">
+          <h4>MediaKeySession destroyed</h4>
+          <p>
+            This algorithm performs steps that are necessary when a <a>MediaKeySession</a> is destroyed without
+            the <a def-id="session-closed-algorithm"></a> algorithm having been run.
+          </p>
+          <p>The following steps are run:</p>
```
ddorwin: Add " in parallel[ to the main thread/event loop]".

CDM operations always run in the parallel to the main thread, including the other instance of the `close the <a href="#key-session">session</a>` text in the `close()` algorithm.

encrypted-media.js
```diff
@@ -233,6 +233,7 @@
     'status-status-pending': { func: idlref_helper, fragment: 'idl-def-MediaKeyStatus.status-pending', link_text: '"status-pending"',  },
 
     'queue-message-algorithm': { func: term_helper, fragment: 'queue-message', link_text: 'Queue a "message" Event',  },
+    'destroy-session-algorithm' : { func: term_helper, fragment: 'destroy-session', link_text: 'Destroy Session', },
```
ddorwin: No longer exists.

jdsmith3000:
I believe all comments on this PR are currently addressed.

encrypted-media-respec.html
```diff
@@ -2310,15 +2310,15 @@
...
-          <h4>MediaKeySession destroyed</h4>
+          <h4>MediaKeySession Destroyed</h4>
           <p>
-            This algorithm performs steps that are necessary when a <a>MediaKeySession</a> is destroyed without
-            the <a def-id="session-closed-algorithm"></a> algorithm having been run.
+            This algorithm performs steps that are necessary when a <a>MediaKeySession</a> is destroyed and
+            the <a def-id="session-closed-algorithm"></a> algorithm has not been run.
           </p>
-          <p>The following steps are run:</p>
+          <p>The following steps are run in parallel [to the main thread/event loop]:</p>
```
ddorwin: Thanks. I meant that the bracketed text was optional, not to actually be included bracketed. If it is included, pick " thread" or "event loop". It looks like we use "event loop" in multiple places. Thus:
>s/[to the main thread/event loop]/to the main event loop/

PR #179, opened 2016-05-03
Alternative PR for Issue #171 with more explicit initiation of close / destroy algorithms

mwatson2:


encrypted-media-respec.html
```diff
@@ -2111,44 +2126,62 @@
...
                 <li>
                   <p>
-                    Destroy these license(s) and/or key(s).
+                    Use <var>cdm</var> to destroy these license(s) and/or key(s).
                   </p>
                 </li>
                 <li>
-                  <p>
-                    If this object's <var>session type</var> is <a def-id="persistent-usage-record-session"></a>, store the <var>record of key usage</var>.
                   </p>
```
jdsmith3000: Replace paragraph close (</p>) with paragraph open (<p>).

encrypted-media-respec.html
```diff
@@ -2111,44 +2126,62 @@
...
-                  <p class="note">
-                    This step may be run asynchronously, including after the document has unloaded.
+                    Run the <a def-id="keys-and-licenses-destroyed-algorithm"></a> algorithm on <var>session</var>.
```
jdsmith3000: Replace </p> above with paragraph open <p>.

jdsmith3000: Replace paragraph close with paragraph open.

jdsmith3000:
The alternative PR splits **Close Session** and **Destroy Keys and Licenses** into two stages, with the second stage called on the last step of the first.  I see also language about using the CDM to "close the session" and "destroy the licenses".  Is the two stage structure necessary to make this emphasis?  From my read, the simpler PR is in some ways clearer, though it doesn't address the explicit processing roles that @ddorwin requests.

mwatson2:
@jdsmith3000 Yes, this is why I kept the first PR and created a new one, since i am not sure either that defining the processing roles more explicitly is necessary (there is no different in observable behavior).

The factoring is the way it is because when a session is destroyed without being closed only a subset of the steps executed on close are needed (specifically, only those parts which do not require interaction with the `MediaKeySession` object which is being destroyed).

jdsmith3000:
@mwatson2 I prefer the first PR.  It's not clear to me that splitting the algorithms is necessary though to clarify the processing roles.  Might your updated language in the two stages still be merged into single stage algorithms?  Or is that necessary to fully satisfy @ddorwin's concerns?

encrypted-media-respec.html
```diff
@@ -1664,12 +1664,7 @@
                             <li><p>Clear all stored session data associated with this object, including the <a def-id="sessionId"></a> and license destruction record.</p>
                               <p class="note">A subsequent call to <a def-id="load"></a> with the value of this object's <a def-id="sessionId"></a> would fail because there is no data stored for that session ID.</p>
                             </li>
-                            <li>
-                              <p>
-                                Run the <a def-id="update-key-statuses-algorithm"></a> algorithm on the <var>session</var>, providing an empty sequence.
```
ddorwin: Good change.  We should make this regardless of which PR, possibly as an independent change along with some of the other cleanup.

mwatson2:
@jdsmith3000 I think the core of @ddorwin's comment was:
> It doesn't make sense for the CDM to call back into the user agent only to have it call back to destroy the keys.

That is, for the case where session closure is initiated by the CDM, we need a procedure to run in the UA that does the clean-up / page interaction but which does not try to destroy the keys (the CDM should already have done that). This is the "Session Closed" algorithm.

And then we also need an algorithm run by the UA which **does** (use the CDM to) destroy the keys, to be called explicitly in `close()` and _Close all Sessions_. This is the "Close Session" algorithm.

Finally, there is a UA-initiated procedure that (use the CDM to) destroy the keys when the session ends without being explicitly closed in one of the two ways described above.

So, I think we do need to split into at least these three, although we could avoid having "Close Session" if we duplicated CDM interaction part in `close()` and _Close all Sessions_.

Finally, in the alternative PR I broke out the "Keys and licenses destroyed" part, just because otherwise that is duplicated in quite a few places. Again, it's only 1 or 2 steps, so it could be left inline.

mwatson2:
This one was useful as an exploration, but I'm not especially happy with it and neither were @jdsmith3000 or (presumably) @ddorwin, so I will close it.

Issue #181, opened 2016-05-06
Fix algorithm inconsistencies


ddorwin:
This issue tracks a set of issues that affect spec and algorithmic correctness and readability. (Many in this initial list were discussed and/or identified in #171 and related PRs but are independent of or should be resolved before #171.) Please feel free to propose related additional candidates in comments. To facilitate more effective reviews, please submit focused PRs for one or more related issue and/or algorithm.

1. #153
1. Clearly document which entity (user agent or CDM) executes each algorithm to reduce confusion and the potential for future issues. (This should be implemented along with #153.)
 * It may also be helpful to specify which entity calls an algorithm where appropriate.
 * For example, "This algorithm is [called by the CDM and] executed in the user agent."
1. Ensure the spec's use of algorithms is consistent with the above changes.
1. Make the Close All Sessions algorithm consistent with the `close()` algorithm (see https://github.com/w3c/encrypted-media/issues/171#issuecomment-217001207).
 * The Close All Sessions algorithm is currently called by both the user agent and CDM.
 * Other than the one case mentioned in the above comment, the Session Close algorithm is only called by the CDM.
1. Be consistent in how the user agent uses the CDM to perform operations (https://github.com/w3c/encrypted-media/issues/171#issuecomment-217001207).
1. Remove the redundant call to `update-key-statuses-algorithm` (https://github.com/w3c/encrypted-media/pull/179#discussion-diff-62074223).
1. Ensure the meaning of "the Session Close algorithm has been run on this object" is unambiguous.
 * Since this can be triggered by various paths and the algorithm itself may (TBD whether it does after fixing the above issues) involve asynchronous operations, it could be unambiguous as to at which point algorithm steps containing the condition above should be triggered.

<br/>Related but not specifically covered by this issue since it involves modifying behavior:
- Address the redundancy between `remove()` and `close()` described in #180.

mwatson2:
I think we need to be careful with language such as "called by the CDM" or "executed by the CDM", since those phrases include execution model assumptions which may not hold in all implementations.

Clearly, there is functionality that is the responsibility of the CDM (including, at least, all the stuff we do not specify such as message formats). For any functionality that we _do_ specify, what does it really mean for us to assign that to the CDM vs the User Agent ? In any given implementation there are likely at least two software components, one of which contains the unspecified CDM functionality. However, the share of specified functionality in the same software component as the unspecified stuff may differ from implementation to implementation.

[...]

Issue #180, opened 2016-05-05
"persistent-usage-record" sessions should not be handled in remove()


ddorwin:
For most ("happy case") `"persistent-usage-record"` sessions, it will be irrelevant whether usage data was persisted. (Conceptually, from the application's point of view, the record is only stored if it was not handled and cleared during the application’s lifetime.) Thus, applications may find it confusing to call `remove()` when they have not explicitly persisted something. Additionally, unlike `"persistent-license"` sessions, `remove()` never actually _removes_ anything for  `"persistent-usage-record"` sessions. (`update()` removes the relevant record for both types of sessions.)

Pulling `"persistent-usage-record"` out of the `remove()` algorithm will focus it on removing (destroying) persisted license(s) and keys(s). (Currently, this is applicable only to `"persistent-license"` sessions. In-memory keys are destroyed when a session is closed, so there is no reason to cover those in this method.) This would make resolving #155 much more straightforward.

<br/>Let’s take a look at what `remove()` currently does for `"persistent-usage-record"` sessions. Currently, it destroys the keys, persists the _record of key usage_, and generates a message.

As the discussion in Issue #171 and related PRs has helped highlight, there are multiple paths to closing sessions and the resulting actions of destroying ephemeral keys and storing the _record of key usage_ for `"persistent-usage-record"` sessions. This seems to make handling this session type in `remove()` redundant except for the queuing of a message. (The in-memory keys are destroyed and the _record of key usage_ is persisted in both cases.)

The session closing paths do not currently perform two operations:

1. Generating a release message when the session is closed.
1. Generating a message for a loaded session containing a persisted record of key usage.
 * This is for the "unhappy case" where the first message was not sent and/or ACK’d.
 * Note that in this case the keys are unnecessarily destroyed (there aren’t any) and the _record of key usage_ is re-stored.

Is there a reason (1) shouldn’t be performed by `close()` (or any other reason the session is closed)?

I see three ways that (2) could be handled. In order of preference:

* `close()` always generates a message for the _record of key usage_, which would have been loaded into the session by `load()`.
 - This seems the most consistent from an application point of view.
 - The loaded sessions is the same as an active session except there are no keys.
* `load()` generates the message, which is something the algorithm already allows.
 - We can add normative steps for this session type there
* `remove()` only generates messages in case (2).
 - This seems inconsistent and makes it hard to define the purpose of `remove()` (see #155).

[...]

PR #184, opened 2016-05-11
PR for issue #181: Session Close refactor

mwatson2:
Refactor the Session Close algorithm into Close Session and Session Closed.


encrypted-media-respec.html
```diff
@@ -318,7 +318,7 @@
           </p>
           <p>Each Key Session is associated with a single <a>MediaKeys</a> object, and only media element(s) associated with that object may access key(s) associated with the session.
           Other <a>MediaKeys</a> objects, <a def-id="cdm"></a> instances, and media elements MUST NOT access the key session or use its key(s).
-          Key sessions and the keys they contain are no longer <a def-id="usable-for-decryption"></a> when the <a def-id="session-close-algorithm">session is closed</a>, including when the <a>MediaKeySession</a> object is destroyed.
+          Key sessions and the keys they contain are no longer <a def-id="usable-for-decryption"></a> when the <a def-id="session-closed-algorithm">session is closed</a>, including when the <a>MediaKeySession</a> object is destroyed.
```
ddorwin: The session-closed-algorithm reverence doesn't seem strictly correct since the keys are no longer usable before that algorithm is called. It might be more accurate to say "once the session has been closed". Note that the referenced algorithm is run "after a session has been closed."

encrypted-media-respec.html
```diff
@@ -2111,20 +2099,44 @@
...
-            <li><p>Run the <a def-id="update-key-statuses-algorithm"></a> algorithm on the <var>session</var>, providing an empty sequence.</p></li>
-            <li><p>Run the <a def-id="update-expiration-algorithm"></a> algorithm on the <var>session</var>, providing <code>NaN</code>.</p></li>
-            <li><p>Let <var>promise</var> be the <a def-id="closed"></a> attribute of the <var>session</var>.</p></li>
-            <li><p>Resolve <var>promise</var>.</p></li>
+            <li><p>Run the <a def-id="session-closed-algorithm"></a> algorithm on the <var>session</var>.</p></li>
           </ol>
         </section>
       </section>
+      
+      <section id="session-closed">
```
ddorwin: Do you intend for this algorithm to be at a higher level than all the other algorithms? I think line 2107 needs to be moved below it.

encrypted-media-respec.html
```diff
@@ -1664,12 +1664,7 @@
                             <li><p>Clear all stored session data associated with this object, including the <a def-id="sessionId"></a> and license destruction record.</p>
                               <p class="note">A subsequent call to <a def-id="load"></a> with the value of this object's <a def-id="sessionId"></a> would fail because there is no data stored for that session ID.</p>
                             </li>
-                            <li>
-                              <p>
-                                Run the <a def-id="update-key-statuses-algorithm"></a> algorithm on the <var>session</var>, providing an empty sequence.
-                              </p>
-                            </li>
-                            <li><p>Run the <a def-id="session-close-algorithm"></a> algorithm on this object.</p></li>
+                            <li><p>Run the <a def-id="close-session-algorithm"></a> algorithm on this object.</p></li>
```
ddorwin: I think these should be session-closed. The keys have already been destroyed, and all other data was removed in the previous step.

encrypted-media-respec.html
mwatson2: Yes. We treat a "session" and a "MediaKeySession" as distinct things, the latter being the "embodiment" of the former according to the Definitions section.

encrypted-media-respec.html
```diff
@@ -1338,7 +1338,7 @@
           <p>The following step is run:</p>
           <ol>
             <li>
-              <p>For each <a>MediaKeySession</a> created by the <var>media keys</var> on which the <a def-id="session-close-algorithm"></a> algorithm has not been run, run the <a def-id="session-close-algorithm"></a> algorithm.</p>
+              <p>For each <a>MediaKeySession</a> created by the <var>media keys</var> on which the <a def-id="session-closed-algorithm"></a> algorithm has not been run, run the <a def-id="close-session-algorithm"></a> algorithm.</p>
```
ddorwin: The Close All Sessions algorithm appears to _only_ be run when the CDM is unavailable. Two things:

1. We should rename this to "CDM Unavailable algorithm" or something like that to make this clear.
1. It doesn't make sense to call Close Session, which uses the CDM to destroy keys if the CDM is unavailable. It seems we should call Session Closed instead.
 * We might want to be clear that the MediaKey
algorithm" or something like that to make this clear.
1. It doesn't make sense to call Close Session, which uses the CDM to destroy keys if the CDM is unavailable. It seems we should call Session Closed instead.
 * We might want to be clear that the MediaKeys is unusable and thus any keys (i.e. in memory) are inaccessible to the app.

encrypted-media-respec.html
mwatson2: I left this as _tbd_ under one of the other items in the OP.

There is a notion of "_closing a session_" (as distinct from closing a `MediaKeySession`) and it is also distinct from "_clearing all session data_". _Closing a session_ does not remove the license destruction record or record of key usage and _clearly all session data_ does not _close the session_ (or, at least, we do not say it does). Here, I think we want to clear all the session data and then _close the session_ and then perform the _Session Closed_ steps. _Close Session_ does exactly those last two things.

encrypted-media-respec.html
```diff
@@ -2089,15 +2080,12 @@
           </ol>
         </section>
 
-        <section id="session-close">
-          <h4>Session Close</h4>
-          <p>The Session Close algorithm is run when the CDM closes the session associated with a <a>MediaKeySession</a> object.</p>
+        <section id="close-session">
```
ddorwin: I expected that `close()` would call this algorithm. Otherwise, I think it only applies to Close All Sessions. It seems those two things should be consistent (except where the CDM is unavailable).

However, see my comments at lines 1341 and 1667. I think we only need to perform these operations in `close()`, meaning we could eliminate this algorithm.

encrypted-media-respec.html
mwatson2: See my [comment on the Issue](https://github.com/w3c/encrypted-media/issues/181#issuecomment-218508687).

encrypted-media-respec.html
ddorwin: Yes, you agree with my suggestion? I don't follow how the second sentence applies here.

Looking at this again, the Session Closed algorithm is _not_ run `including when the <a>MediaKeySession</a> object is destroyed` because the object no longer exists. Either we should remove the session-closed-algorithm link or convert "including" to "or" or similar.

encrypted-media-respec.html
```diff
@@ -1724,11 +1714,12 @@
           <p class="note">The returned promise is resolved when the request has been processed, and the <a def-id="closed"></a> attribute promise is resolved when the session is closed.</p>
 
           <ol class="method-algorithm">
-            <li><p>If the <a def-id="session-close-algorithm"></a> algorithm has been run on this object, return a resolved promise.</p></li>
+            <li><p>If the <a def-id="session-closed-algorithm"></a> algorithm has been run on this object, return a resolved promise.</p></li>
             <li><p>If this object's <var>callable</var> value is false, return a promise rejected with an <a def-id="InvalidStateError"></a>.</p></li>
             <li><p>Let <var>promise</var> be a new promise.</p></li>
             <li><p>Run the following steps in parallel:</p>
               <ol>
+                <!--- Consider whether to replace the following with a call to Close Session -->
```
ddorwin: See my comment at line 2083.

If this does remain unresolved in this PR (I don't think it should), add "TODO:" so we don't forget about it.

encrypted-media-respec.html
mwatson2: Ok, I will roll in a couple of the other issues into this PR. Specifically, make things consistent in terms of "use the CDM to..." and have a small and consistent set of things that we use the CDM to do with respect to the session:
- close the session (which destroys all non-persistent licenses and keys)
- destroy persisted licenses and keys
- destroy all persisted state (record of license destruction, record of key usage)

encrypted-media-respec.html
mwatson2: Yes, I agree with your suggestion. I was just noting that it makes sense to refer to a "session" as distinct from a `MediaKeySession`.

I'll remove the link. Then I think the _including_ is correct.

encrypted-media-respec.html
```diff
@@ -2195,7 +2207,7 @@
         </p>
         <p>The CDM MUST ensure that data for a given session is only present in one active unclosed session in any <a def-id="document-concept"></a>.
           In other words, <a def-id="load"></a> MUST fail when there is already a <a>MediaKeySession</a> representing the session specified by the <var>sessionId</var> parameter, either because the object that created it via <a def-id="generateRequest"></a> is still active or it has been loaded into another object via <a def-id="load"></a>.
-          A session MAY only be loaded again after the <a def-id="session-close-algorithm"></a> algorithm has not been run on the object representing it.
+          A session MAY only be loaded again after the <a def-id="session-closed-algorithm"></a> algorithm has not been run on the object representing it.
```
ddorwin: Line 2208:
We use "unclosed session" three times in the spec, but this term is not defined. Should we a) replace these with `the <a def-id="session-closed-algorithm"></a> algorithm has been run on this object` or b) add a definition for closed session, possibly using this text, and update all such text as well as these three cases to reference the definition?

(b) might be better from a readability perspective and avoids having to ensure we get the somewhat complex text correct everywhere.

encrypted-media-respec.html
ddorwin: Also, re-indenting the text will make the diff a bit smaller.

encrypted-media-respec.html
```diff
@@ -2111,20 +2099,44 @@
...
+      <section id="session-closed">
+        <h4>Session Closed</h4>
+        <p>The Session Closed algorithm updates the <a>MediaKeySession</a> state after a session has been closed.</p>
+        </p>
+        <div class="note">
+          <p>
+            The CDM may close a session at any point, such as when the session is no longer needed or when system resources are lost.
+            In that case, this algorithm MUST be run.
+          </p>
+          <p>Keys in other sessions SHOULD be unaffected, even if they have overlapping key IDs.</p>
```
ddorwin: Unrelated to this change, it seems this `SHOULD` should be `MUST`. Otherwise, there is a difference in observable behavior. At a minimum, non-overlapping key IDs should be unaffected, but I think the spec should expect correct and consistent behavior unless there is a good reason not to.

mwatson2: I agree, but perhaps we should file a separate Issue for this in case anyone else especially cares.

encrypted-media-respec.html
mwatson2: I think we will need separate definitions for _closed session_ and _closed `MediaKeySession`_. The latter implies the former, but not the other way around, first because the session is closed just before the `MediaKeySession` in the normal case and second because `MediaKeySession` can be destroyed without being closed, but the _session_ is still closed in this case.

ddorwin: I filed #185 to track this since it's unrelated.

encrypted-media-respec.html
```diff
@@ -1331,14 +1331,16 @@
...
         </section>
         
-        <section id="close-all-sessions">
-          <h4>Close all sessions</h4>
-          <p>The close all sessions algorithm is run to close all sessions associated with a <a>MediaKeys</a> object, <var>media keys</var>.
+        <section id="cdm-unavailable">
+          <h4>CDM Unavailable</h4>
+          <p>
+            The CDM unavailable algorithm is run to close all <a>MediaKeySession</a> objects associated with a <a>MediaKeys</a> object, <var>media keys</var>
+            when the CDM becomes unavailable.
```
ddorwin: nit: "when its CDM instance..." or something similar.

encrypted-media-respec.html
```diff
@@ -1661,30 +1663,28 @@
                         <dd>
                           <p>Run the following steps:</p>
                           <ol>
-                            <li><p>Clear all stored session data associated with this object, including the <a def-id="sessionId"></a> and license destruction record.</p>
-                              <p class="note">A subsequent call to <a def-id="load"></a> with the value of this object's <a def-id="sessionId"></a> would fail because there is no data stored for that session ID.</p>
-                            </li>
                             <li>
                               <p>
-                                Run the <a def-id="update-key-statuses-algorithm"></a> algorithm on the <var>session</var>, providing an empty sequence.
+                                Use <var>cdm</var> to clear all stored session data associated with this object, including the <a def-id="sessionId"></a>
```
ddorwin: This step is already running in the CDM ("Use the cdm to execute the following steps:" above).

mwatson2: Yes, and there are also quite a few things in this section which are User Agent procedures (closing the session, updating status, updating expiration, sending a message).

One approach would be to remove all those things and call the "Monitor CDM Changes" algorithm, which would then queue all the appropriate events before the `update()` method promise was resolved.

encrypted-media-respec.html
```diff
@@ -2089,34 +2078,48 @@
...
             All <a>MediaKeySession</a> methods will fail for this object.
           </p>
-          <p class="note">The CDM may close a session at any point, such as in response to a <a def-id="close"></a> call, when the session is no longer needed, or when system resources are lost.
-          Keys in other sessions SHOULD be unaffected, even if they have overlapping key IDs.
+          <p>The following steps are run:</p>
+          <ol>
+            <li><p>Let <var>session</var> be the associated <a>MediaKeySession</a> object.</p></li>
+            <li><p>Let <var>cdm</var> be the CDM instance represented by <var>session</var>'s <var>cdm instance</var> value.</p></li>
+            <li>
+              <p>Use <var>cdm</var> to close the <a href="#key-session">session</a> associated with <var>session</var>.</p>
```
ddorwin: Previously:

1. `close()` initiated a request to close the session
1. The session may not have been closed.
1. There was normative text to "not remove stored session data. Do not generate or send key or license release messages."

I don't recall the reasons for (1) and (2), and it seems reasonable to require that the session be close. (3) seems important. See also the comment on the next line.

encrypted-media-respec.html
```diff
@@ -2089,34 +2078,48 @@
...
+              <p class="note">Closing the key session results in the destruction of any license(s) and key(s) that have not been explicitly stored.</p>
```
ddorwin: Similar to (3) above, previously, there was a normative step to "Destroy these license(s) and/or key(s)" (in the Session Close algorithm). Is there a reason this has been replaced with a Note? IOW, why wouldn't we specify that here?

encrypted-media-respec.html
ddorwin: >Yes, and there are also quite a few things in this section which are User Agent procedures (closing the session, updating status, updating expiration, sending a message).

I think we should fix these as part of #181 (items 2 and 3). For now, though, I don't think we should add redundant text to use the cdm to use the cdm to do something.

>One approach would be to remove all those things and call the "Monitor CDM Changes" algorithm, which would then queue all the appropriate events before the `update()` method promise was resolved.  Or we could just pull them out of the "Use the cdm to..." section.

I think we can resolve this without completely removing these steps. For example, we can have a _should close_/_is closed_ variable then run the appropriate algorithm in the UA portion of the steps like how we handle _message_.

encrypted-media-respec.html
```diff
@@ -1661,30 +1663,28 @@
...
                             <li>
                               <p>
-                                Run the <a def-id="update-key-statuses-algorithm"></a> algorithm on the <var>session</var>, providing an empty sequence.
+                                Use <var>cdm</var> to clear all stored session data associated with this object, including the <a def-id="sessionId"></a>
+                                and license destruction record.
                               </p>
+                              <p class="note">A subsequent call to <a def-id="load"></a> with the value of this object's <a def-id="sessionId"></a> would fail because there is no data stored for that session ID.</p>
                             </li>
-                            <li><p>Run the <a def-id="session-close-algorithm"></a> algorithm on this object.</p></li>
+                            <li><p>Run the <a def-id="close-session-algorithm"></a> algorithm on this object.</p></li>
```
ddorwin: In response to https://github.com/w3c/encrypted-media/pull/184/commits/251debc2d52c9a0202bfa8e7e26b16107def8d2f#r62881880:

What does it mean to close a session other than destroy the keys or removing session data (at least from an observable point of view)? I do not think it should be possible to get to this point without having gone through normative steps that closed the session but did not remove the session data. Thus, this seems redundant, especially since going back through the Close Session algorithm is confusing. (While not critical, it also adds another algorithm that is called both from the CDM and UA portion of algorithms.)

encrypted-media-respec.html
ddorwin: SGTM. Filed #186.

encrypted-media-respec.html
ddorwin: To be clearer about my request:

1. Remove the redundant "use the cdm" text in this step.
1. We'll clean up the rest in a separate PR, possibly as discussed above.

ddorwin:
Looks good overall. I think there are only a few remaining issues for this PR.

encrypted-media-respec.html
mwatson2: Agreed - although I already, independently, made the change as you suggest above in my local copy of the PR. I think it's sufficiently related to do it together, but I could pull it out if you feel strongly.

encrypted-media-respec.html
mwatson2: > What does it mean to close a session other than destroy the keys or removing session data (at least from an observable point of view)?

This is a good question. We could say that "closing the session" is identical to "destroying all licenses, keys and other session data that have not been explicitly stored" and then we could consistently use one phrase or the other. Using only the second phrase is inconvenient for the case where it is the CDM that "closed the session".

In this particular algorithm, though, it is *all* session data that is being destroyed, including that which has been explicitly stored. That is a superset of ""destroying all licenses, keys and other session data that have not been explicitly stored" and so if we agree on the above identity then you are right that this closes the session. But this would not be very clear to the reader.

Perhaps we should say above "Close the session and destroy any data which has been stored, including the license destruction record". The  we will call session-closed and perhaps we don't need Close Session at all ?

encrypted-media-respec.html
mwatson2: Per the discussion above, "Closing the session" is either identical to or perhaps a superset of "Destroy all licenses and keys and session data that has not been explicitly stored". We should certainly specify that somewhere (I'll do a sweep and ensure we do) and then use the "Close the session" phrase consistently.

encrypted-media-respec.html
mwatson2: I deliberately dropped the possibility that the session doesn't get closed. Without a clear use-case it seems sensible that `close()` should always result in the session closing. We might revisit this depending on the solution to #180.

mwatson2: Regarding (3), it seemed to me that the procedures should be sufficiently clear about what closing the session involves that we don't need these clarifications. But I will check into this again.

encrypted-media-respec.html
mwatson2: There are other examples where algorithm invocations occur within "Use _cdm_ to ..." blocks, so I will leave that aspect for another PR.

encrypted-media-respec.html
mwatson2: I implemented the above suggestion, removing Close Session (since it was only used by `close()`) and adding a normative requirement to the Key Session definition that the licenses and keys are deleted when the session is closed.

encrypted-media-respec.html
mwatson2: (3) should now be clear in the Key Session definition and the `close()` method.

encrypted-media-respec.html
```diff
@@ -2077,28 +2086,13 @@
...
-              <p class="note">Closing the key session results in the destruction of any license(s) and key(s) that have not been explicitly stored.</p>
-            </li>
-            <li><p>Run the <a def-id="session-closed-algorithm"></a> algorithm on the <var>session</var>.</p></li>
-          </ol>
-        </section>
       
         <section id="session-closed">
           <h4>Session Closed</h4>
           <p>The Session Closed algorithm updates the <a>MediaKeySession</a> state after a <a href="#key-session">session</a> has been closed.</p>
+          <p>
```
ddorwin: Similar to an earlier comment (https://github.com/w3c/encrypted-media/pull/184/#discussion_r62946959), it seems like it would be useful to at least note that release message(s) are not sent here (at least pending #180). Although we explicitly do not allow for it, an implementation could send one using the busy-loop algorithm being discussed in #181.

Really, this could apply to any place that the session is closed, but this is the algorithm that will _always_ be called. We could word the note to cover all such cases rather than just this algorithm. Perhaps this would be a good second paragraph in the note div below

encrypted-media-respec.html
```diff
@@ -2088,35 +2086,34 @@
...
+            Closing a <a href="#key-session">session</a> means that the license(s) and key(s) associated with it are no longer available to
+            decrypt <a def-id="media-data"></a>. All <a>MediaKeySession</a> methods will fail for this object after this algorithm is run.
           </p>
-          <p class="note">The CDM may close a session at any point, such as in response to a <a def-id="close"></a> call, when the session is no longer needed, or when system resources are lost.
-          Keys in other sessions SHOULD be unaffected, even if they have overlapping key IDs.
+          <div class="note">
+            <p>
+              The CDM may close a session at any point, such as when the session is no longer needed or when system resources are lost.
+              In that case, this algorithm MUST be run.
+            </p>
```
ddorwin: Similar to an earlier comment (https://github.com/w3c/encrypted-media/pull/184/#discussion_r62946959), it seems like it would be useful to at least note that release message(s) are not sent here (at least pending #180). Although we explicitly have a step for it, an implementation could send one using the busy-loop algorithm being discussed in #181.

Really, this could apply to any place that the session is closed, but this is the algorithm that will _always_ be called. We could word the note to cover all such cases rather than just this algorithm. Placing it after the above statement about when to run this step might also help.

ddorwin:
LGTM with one comment. Thank you!

encrypted-media-respec.html
mwatson2: [Hmm, I thought I posted this comment already - apologies for duplicates]

I will add a note.

I assume the busy loop algorithm runs in the same "thread" (the one-and-only thread in which web platform specification algorithms execute) and so that algorithm cannot be used to send a message or anything because it only runs for as long as the session is not closed.

mwatson2:
Merged in 226d811a9fb2706c06ef00012a61bf056b839377

Issue #31, opened 2015-02-10
generateRequest() should allow the first message to not be a license request based on initData


ddorwin:
The generate license request algorithm was always intended to support multiple messages/round trips. This could be part of the protocol or for other things, such as to obtain a server certificate when one has not been provided otherwise and an `individualization-request` message (see issue #19). The algorithm (now `generateRequest()`) has evolved to require the first message be a license request generated from the initData.

We should change the algorithm to allow such other types of messages while still ensuring that a license request is generated (in the absence of errors) following the existing algorithm. This probably involves breaking that algorithm out and specifying when to call it from the `generateRequest()` and `update()` algorithms.

We need to address #19 first so that we can make these changes in a way that ensures the correct ordering.

ddorwin:
Since #19 seeks to ensure that events are not fired until after the promise is resolved, we will need to ensure that these changes maintain that. For example, we may need to ensure that initData, etc. are processed (and perhaps the request partially generated) before the promise is resolved so that any errors can be reported by rejecting the returned promise.

Thus, rather than extracting the generation algorithm, we may just need to save and restore its state.

cpearce:
Yes, sometimes we will need the first message to be an indiv request, and subsequent messages may be required until a license request can be sent.

If multiple messages can be sent after the promise returned by a generateRequest() call is resolved, we'll need a way to report errors that result from the responses from those errors. I'm not sure if reporting errors by rejecting the promises returned by subsequent update() calls (as you proposed in issue #14 )  is sufficient. I suspect it is, as the only thing not covered by this would be a network timeout, but that can be handled by the JS layer.

steelejoe:
I don't believe this should be blocked by issue #19. I do believe this is blocked by issue #41. 

If issue #41 is resolved to allow for not sending requests when they are not required by the CDM, then the text change is very simple:

`Generates a license request based on the initData. A message of type "individualization-request" or  "license-request"  may be queued if the algorithm succeeds and the promise is resolved.`

If issue #41 is deferred or resolved to continue requiring requests to always be generated, then the text is also very simple:

`Generates a license request based on the initData. A message of type "individualization-request" or  "license-request"  will be queued if the algorithm succeeds and the promise is resolved.`

In either case the `update` 6.6.5.2 algorithm must change to reflect the new message type added. I believ we should replace the current text:

`Let message type be the appropriate MediaKeyMessageType for the message.`

With this:

[...]

PR #190, opened 2016-05-13
PR for issue #31: Allow generateRequest() to send an individualization request

mwatson2:


encrypted-media-respec.html
```diff
@@ -1542,17 +1542,27 @@
...
                     </li>
-                    <li><p>Let <var>message</var> be a license request for the <var>requested license type</var> generated based on the <var>sanitized init data</var>, which is interpreted per <var>initDataType</var>.</p>
+                    <li><p>Let <var>message</var> be a license request for the <var>requested license type</var> generated based on the <var>sanitized init data</var> interpreted per <var>initDataType</var>, or an individualization request.</p>
                       <p>The <var>cdm</var> MUST NOT use any stream-specific data, including <a def-id="media-data"></a>, not provided via the <var>sanitized init data</var>.</p>
                       <p>The <var>cdm</var> SHOULD NOT store session data, including the session ID, at this point. See <a def-id="session-storage"></a>.</p>
+                      <o>If <var>message</var> is an individuallization request, then in a subsequent call to <a def-id="update"></a> the CDM MUST generate a license request for the <var>requested license type</var>  based on the <var>sanitized init data</var>, which is interpreted per <var>initDataType</var>.</p>
```
mwatson2: s/<o>/<p>/ and s/individuallization/individualization/

mwatson2: s/is an individualization request/is not a license request/

encrypted-media-respec.html
```diff
@@ -1542,17 +1542,27 @@
                     <li><p>Let <var>session id</var> be a unique <a def-id="session-id"></a> string.</p>
                       <p>If the result of running the <a def-id="is-persistent-session-type-algorithm"></a> algorithm on <var>session type</var> is <code>true</code>, the ID MUST be unique within the <a def-id="origin"></a> of this object's <a def-id="document-concept"></a> over time, including across Documents and browsing sessions.</p>
                     </li>
-                    <li><p>Let <var>message</var> be a license request for the <var>requested license type</var> generated based on the <var>sanitized init data</var>, which is interpreted per <var>initDataType</var>.</p>
+                    <li><p>Let <var>message</var> be a license request for the <var>requested license type</var> generated based on the <var>sanitized init data</var> interpreted per <var>initDataType</var>, or an individualization request.</p>
```
ddorwin: This does not cover all use cases. Specifically, it doesn't cover the server certificate request. I think we need something more general that results in the requested license request or not.

I think we can cover all three known cases (license request, server certificate, individualization) by checking whether a license request can be generated at this time. Even the fourth theoretical case of a multi-round license request process could probably fall into a) generating some type of request or b) not being able to at this time.

For example:
* If a license request can be generated at this time, let _message_...{existing text}
* Otherwise:
   - Run the following steps:
      1. Let _message_ be a request that needs to be processed before generating a request.
      1. Let _message type_ be `<a def-id="message-type-individualization-request">` for xyz and `<a def-id="message-type-license-request"></a>` otherwise.
   - When the response to _message_ is provided in a subsequent call to `<a def-id="update"></a>`, the CDM MUST... {text below}<br/>{Issue box from below}

encrypted-media-respec.html
```diff
@@ -1542,17 +1542,27 @@
...
                       <p>The <var>cdm</var> MUST NOT use any stream-specific data, including <a def-id="media-data"></a>, not provided via the <var>sanitized init data</var>.</p>
                       <p>The <var>cdm</var> SHOULD NOT store session data, including the session ID, at this point. See <a def-id="session-storage"></a>.</p>
+                      <o>If <var>message</var> is an individuallization request, then in a subsequent call to <a def-id="update"></a> the CDM MUST generate a license request for the <var>requested license type</var>  based on the <var>sanitized init data</var>, which is interpreted per <var>initDataType</var>.</p>
+                      <p class="issue">
+                        <a href="https://github.com/w3c/encrypted-media/issues/31">Issue 31</a> proposes that the above requirement be formally
```
ddorwin: I filed #207 to specifically track future work. We can use that number here and close #31 with this PR.

encrypted-media-respec.html
ddorwin: I'm open to other solutions as well. I like that this one clearly breaks out the two cases (request generated and not) and allows the `update()` requirement to be in the second block.

Issue #158, opened 2016-04-12
Privacy: Prohibit access/use of sensitive data (e.g. location) by CDMs


ddorwin:
In https://github.com/w3c/encrypted-media/issues/157#issuecomment-208844577, @mwatson2 says:
>For online viewing, services may indeed apply geographic restrictions. ...it is a server function to apply these restrictions, not something that is done by the DRM. This is important to recognize because there would be privacy implications if the CDM could access your location.

While we assume the CDM cannot access or use the client's/user's location, I'm not sure it is currently expressly prohibited by the spec.

More generally, the CDM should not use (have access to?) or expose data that is not generally available to web applications or is generally protected by a user permission and/or prompt. Location is a primary example, but there are others, both exposed to the web (i.e. user media, such as camera and mic) and not (i.e. LAN details or devices).

While the examples above may seem clear cut, the phrasing could be tricky, especially since unsandboxed CDMs often do have such access and some CDMs use, for example, Distinctive Identifiers not otherwise exposed.

Note that preventing exposure of such data is not sufficient since even use of them could allow them to be derived (i.e. via a series of licenses).

mwatson2:
In my statement I was assuming that the text around privacy and distinctive identifiers excluded the exposure by the CDM of personally identifiable information generally, but perhaps we need to review tighten the text there.

In general, when we talk about restricting "exposure" of information we should include within this "indirect exposure" (i.e. ability to derive information over time). CDMs may "use" various kinds of information, for example to make the identifiers they use distinctive ( = not shared by other users / devices), but the important point is that any exposed information can be reset to prevent tracking over time. Put another way, no time-persistent / non-resettable information should be exposed (directly or indirectly).

josephlhall:
Hi, Joe Hall from W3C's Privacy Interest Group (PING) here (also a lurker in EME/public-html-media).

It sounds like EME calls and even CDMs don't need to request permissions through the browser for access to certain sensitive data like those called out in the [Secure Contexts](https://w3c.github.io/webappsec-secure-contexts/#threat-risks) draft?

PING would be happy to discuss this further, although it sounds like from what @mwatson2 says near the end there are limited privacy implications.

steelejoe:
[...]

PR #199, opened 2016-05-23
Issue #158: Explicitly prohibit expose of unspecified information thr…

mwatson2:
…ough CDM messages

### DIFF
--- a/encrypted-media-respec.html
+++ b/encrypted-media-respec.html
@@ -3302,8 +3302,9 @@
             <li><p>Only data stored by the requesting <a def-id="origin"></a> may be loaded.</p></li>
             <li>
               <p>
-                It is not possible to extract, derive or infer information from CDM messages that is not either explicitly described in this specification
-                or available to the page through other web platform APIs.
+                It is not possible to extract, derive or infer information from the CDM that is not either explicitly described in this specification
+                or available to the page through other web platform APIs without user permission. This applies to any information that is exposed outside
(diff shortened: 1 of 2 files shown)


encrypted-media-respec.html
```diff
@@ -3218,6 +3218,23 @@
             <li><p>MediaKeys objects and the underlying implementation do not expose information outside the <a def-id="origin"></a>.</p></li>
             <li><p>Persisted session data, if applicable, is stored on a per-<a def-id="origin"></a> basis.</p></li>
             <li><p>Only data stored by the requesting <a def-id="origin"></a> may be loaded.</p></li>
+            <li>
+              <p>
+                It is not possible to extract, derive or infer information from CDM messages that is not either explicitly described in this specification
```
ddorwin: s/is not/MUST NOT be/ ?

encrypted-media-respec.html
```diff
@@ -3218,6 +3218,23 @@
...
+                or available to the page through other web platform APIs.
```
mwatson2: add "... without user permission."

encrypted-media-respec.html
ddorwin: Oops. This is addressed in 3210.

ddorwin: Rather than saying "CDM messages," which could be interepreted as `message` events, we might want to use language similar to Distinctive Identifier: "It is exposed outside the client device or exposed to the application..."

PR #200, opened 2016-05-23
Issue #129: Require readyState transition on key unavailable / available

mwatson2:


encrypted-media-respec.html
```diff
@@ -2657,6 +2658,16 @@
                   <p>Set the <var>media element</var>'s <var>waiting for key</var> value to <code>false</code>.</p>
                   <p class="note">As a result of the above step, the media element may no longer be a <a def-id="blocked-media-element"></a> and thus playback may resume.</p>
                 </li>
+                <li>
+                  <p>
+                    Set the <var>media element</var>'s <a def-id="readystate"></a> value to <a def-id="have-future-data"></a> or
```
ddorwin: This can probably be `have-current-data` here too.

encrypted-media-respec.html
```diff
@@ -2657,6 +2658,16 @@
...
+                    <a def-id="have-enough-data"></a> as appropriate.
```
mwatson2: Could be HAVE_CURRENT_DATA too.

Link "as appropriate" to the definitions.

encrypted-media-respec.html
```diff
@@ -2657,6 +2658,16 @@
...
+                  </p>
+                  <p class="note">
```
ddorwin: We might also note that the appropriate events for the transition are fired.

Issue #177, opened 2016-04-30
Transfer CDM configuration information to the instance creation in createMediaKeys()


ddorwin:
The results of the `requestMediaKeySystemAccess()` algorithms affect the CDM instance(s) created by `createMediaKeys()` in a few normative ways. For example, whether the instance is allowed to persist state or use Distinctive Identifiers are both normatively determined by the corresponding members of the resulting `MediaKeySystemConfiguration`.

However, I believe the intent is that other decisions made in these algorithms should also affect the CDM instance(s). This is especially relevant for implementation-specific items that require consent (as being discussed in PR #165 and issue #96), but other properties, such as content type and robustness, may also affect the configuration or even the pipeline that is used.

We should explicitly document this transfer of configuration information and restrictions from the `requestMediaKeySystemAccess()` algorithms to the resulting `MediaKeySystemAccess` object and subsequent use by `createMediaKeys()`. Since the type of information is implementation-specific, we can use a symbolic internal variable similar to the <var>restrictions</var> variable in PR #165.

ddorwin:
Are there any comments on this or is it ready for implementation?

mwatson2:
I agree that the entire value of _configuration_ should be used in `createMediaKeys` when creating the CDM _instance_ (Step 7). But I am not sure what other implementation-specific information you are expecting to need ?

Whatever that was would need to be created during `requestMediaKeySystemAccess`, stored with the `MediaKeySystemAccess` object and used in the creation of the CDM _instance_.

Even if there is such information, it's not clear to be that its use or otherwise has any visible affects and so needs to be documented.

mwatson2:
[Here](https://github.com/w3c/encrypted-media/pull/201) is a PR for the _configuration_ part.

ddorwin:
>Even if there is such information, it's not clear to me that its use or otherwise has any visible affects and so needs to be documented.

I was probably thinking about the implementation-specific restrictions or other decisions that might be recorded. However, we are probably fine with just the above change for  now. If there becomes a need to configure based on other information, we can add that later. Also, while implementations may choose to pass around other state, it is probably equivalent to reprocessing the configuration and generating that state in the new initialize step.

In any case, the intent is clear: the returned configuration affects the CDM instance.

PR #201, opened 2016-05-23
Issue #177: Initialize CDM instance using the full configuration

mwatson2:


encrypted-media-respec.html
```diff
@@ -1170,6 +1170,7 @@
                 </li>
                 <li><p>Load and initialize the <a def-id="keysystem"></a> implementation represented by this object's <var>cdm implementation</var> value if necessary.</p></li>
                 <li><p>Let <var>instance</var> be a new instance of the <a def-id="keysystem"></a> implementation represented by this object's <var>cdm implementation</var> value.</p></li>
+                <li><p>Initialize <var>instance</var> using <var>configuration</var>.</p></li>
```
ddorwin: We might be explicit about the meaning of this. For example, that CDM features may be enabled/disabled/selected based on the contents of `<var>configuration</var>`.

ddorwin:
LGTM with suggestion.

jdsmith3000:
Made this change:

> Initialize _instance_ to enable, disable and/or select `Key System` features using _configuration_.

Issue #187, opened 2016-05-12
Inconsistency about requirements to clear persistent data


xhwang-chromium:
In 11.5 Information Stored on User Devices -> 11.5.2 Mitigations, the spec says

> In summary, user agent and CDM implementations that allow the CDM to persist data:
> ...
> **must** allow the user to clear it, **preferably by origin**.
> 

The in "User deletion of Key System storage" section, it says:

> User agents **should** present the user with a way to delete Key System storage **for a specific origin or all origins**.

It becomes a bit confusing as for whether providing a way for user to clear the data is a requirement or a recommendataion. 

Also, it's less clear about whether "by origin" or "for a specific origin" is more preferable, or it is actually equally acceptable as "for all origins".

mwatson2:
> In summary, user agent and CDM implementations that allow the CDM to persist data:
...
MUST allow the user to clear it ...

This seems clear to me. The UA must allow the user to clear the CDM persisted data. Any User Interface option which provided for this would meet the requirement, including options that did much more, such as a non-origin-specific, non-EME-specific "Clear all stored data" option. 

> ... preferably by origin.

"preferably" is not an RFC2119 term, so this could be clarified, but I would interpret this to have "recommendation" strength.

> User agents should present the user with a way to delete Key System storage for a specific origin ..

This may be a repetition of the recommendation above. However the implication here is that the User Interface provides for specifically clearing Key System storage (e.g. "Clear all licenses, keys and other content protection data"). This could be clarified, for example to say "... a _specific_ way to delete ...".

> ... or all origins.

Without the clarification, this is a restatement of the original requirement, but at lower strength, which is indeed confusing. With the clarification above, the recommendation here is that the method of clearing Key System data, whether per origin (recommended) or not, be specific to Key System data and not bundled with clearing of other data.

I suggest we make the clarifications above, but I don't think they are essential for V1.

mwatson2:
Please see this [pull request](https://github.com/w3c/encrypted-media/pull/202).

ddorwin:
As I said in #202, that PR addresses the requirement inconsistency and replaces "preferably," which doesn't have a normative meaning. I filed #210 to track structural issues with this section. I think we can close this issue now.

xhwang-chromium:
Agreed this issue can be closed now. After #210 is addressed we shouldn't have any inconsistency issues.

PR #202, opened 2016-05-23
Issue #187: Clarify requirements for clearing persistent store

mwatson2:


encrypted-media-respec.html
```diff
@@ -3450,7 +3451,7 @@
 
             <dt>User deletion of Key System storage</dt>
             <dd>
-              <p>User agents SHOULD present the user with a way to delete Key System storage for a specific <a def-id="origin"></a> or all origins.
+              <p>User agents SHOULD present the user with a way to delete Key System storage specifically, by <a def-id="origin"></a> or for all origins.
                 User agents SHOULD help the user avoid <a href="#incomplete-clearing">Incomplete Clearing of Data</a>.
```
mwatson2: @ddorwin suggests "to specifically delete ..."

encrypted-media-respec.html
```diff
@@ -3450,7 +3451,7 @@
 
             <dt>User deletion of Key System storage</dt>
             <dd>
-              <p>User agents SHOULD present the user with a way to delete Key System storage for a specific <a def-id="origin"></a> or all origins.
+              <p>User agents SHOULD present the user with a way to delete Key System storage specifically, by <a def-id="origin"></a> or for all origins.
```
mwatson2: @ddorwin suggests "to specifically delete ..."

encrypted-media-respec.html
```diff
@@ -3611,7 +3612,7 @@
 
             <dt>User deletion of Key System storage</dt>
             <dd>
-              <p>User agents SHOULD present the user with a way to delete Key System storage for a specific <a def-id="origin"></a> or all origins.
+              <p>User agents SHOULD present the user with a way to specifically delete Key System storage, by <a def-id="origin"></a> or for all origins.
```
ddorwin: "or" ==> "and/or" ?

jdsmith3000: "or" seems correct.  The way is provided by origin OR for all origins.  "by origin" seems encompassed by "all origins".

ddorwin:
This change LGTM to address the requirement inconsistency and replace "preferably." I filed #210 to track structural issues, which are low priority.

Note that this change introduces a new `SHOULD` for explicitly/specifically exposing the ability to clear Key System storage, not just the `MUST` requirement to allow clearing it (i.e. as part of some "clear everything" mechanism).

encrypted-media-respec.html
ddorwin: I'm open to "or" being sufficient, but I don't think "all origins" encompasses "by origin." It might encompass "for an origin," but "by origin" means one at a time.

For example, "by origin" means the user can choose to delete Key System data just for example.com but not for foo.com and all other origins.

jdsmith3000: My interpretation is "by origin" means for a specific origin.  So as written, this section says that the UA provides a feature for specific selected origins or across all origins.

jdsmith3000: I retract my comment.  "And/or" is preferable, since it allows both.

jdsmith3000:
Updated for minor review feedback.

PR #204, opened 2016-05-26
Fix #178: Require audio and video capabilities with at least one not empty

### LINKED ISSUE #178: Be explicit about how the lack of audioCapabilities/videoCapabilities in a configuration is handled
The [Get Supported Configuration](https://w3c.github.io/encrypted-media/#get-supported-configuration) algorithm currently allows `audioCapabilities` and `videoCapabilities` to not be present in the `MediaKeySystemConfiguration`, but the spec does not specify how this is to be handled/interpreted.

When the capabilities member is present, the sequence must be non-empty and for each entry:
* `contentType` must be non-empty.
* `robustness` may be empty, in which case we rely on this text from the [definition](https://w3c.github.io/encrypted-media/#widl-MediaKeySystemMediaCapability-robustness) of this member: "The empty string indicates that any ability to decrypt and decode the content type is acceptable."

I see two options:

**Option 1: Require these members be present.**
There may not be sufficient reason to allow these members to be optional. Applications _should_ specify them in most cases, especially since this should improve the chances of success across client implementations.

I can think of two scenarios where they legitimately might not need to be specified:

1. Audio-only content: `videoCapabilities` is unnecessary.
1. Managing persistent state: Content type is irrelevant and it's simpler to not specify types to cover all implementations.

(1) can be supported by a) stating that content types not specified are not supported by the instance or b) requiring an explicit indication, such as an empty list. We could additonally prohibit not specifying _both_ members, which would ensure there is always a robustness specified (even if per the existing text regarding the empty string).

It may not be necessary to support (2):
* Even when media data won't be decrypted by the instance, robustness is still relevant as discussed above.
* It is probably better to fully specify the requirements so the `MediaKeySystemAccess` (and `MediaKeys`) object can be reused for playback and to avoid the potential for multiple user consent requests.

**Option 2: Specify the behavior.**
Specify something similar to the text for `robutstness` as the behavior when `audioCapabilities` or `videoCapabilities` - and perhaps both - are not present.

In such cases, there will be no guarantees of any audio or video type support, respectively. Thus, the most important clarification is the robustness level, perhaps especially when neither member is present. (We should consider that robustness could theoretically be relevant even when no content is to be decrypted. As a hypothetical example: case (2) above if there multiple pipelines.)

We could specify this in the two member definitions. Along with `robustness`, I wonder whether we should also normatively specify this in the algorithm steps or non-normatively refer to the definition at the appropriate step in the algorithm.

mwatson2:
…be empty


encrypted-media-respec.html
```diff
@@ -694,21 +694,42 @@
                 </ol>
               </li>
               <li><p>Set the <a def-id="option-sessionTypes"></a> member of <var>accumulated configuration</var> to <var>session types</var>.
+              
+              <li>
+                <p>
+                  If the <a def-id="option-videoCapabilities"></a> and <a def-id="option-audioCapabilities"></a> members in <var>candidate configuration</var>
```
ddorwin: For clarity/emphasis, add "both": "If **both* the..."

ddorwin:
LGTM with suggestion.

ddorwin:
nit: The commit message is a bit confusing. I think it's more of an "and" (or just ';') than "but."

jdsmith3000:
Updated to emphasize that both audio and video capabilities cannot be empty.

encrypted-media-respec.html
```diff
@@ -697,8 +697,8 @@
               
               <li>
                 <p>
-                  If the <a def-id="option-videoCapabilities"></a> and <a def-id="option-audioCapabilities"></a> members in <var>candidate configuration</var>
-                  are empty, return <code>NotSupported</code>.
+                  If both the <a def-id="option-videoCapabilities"></a> and <a def-id="option-audioCapabilities"></a> members in <var>candidate configuration</var>
+                  are both empty, return <code>NotSupported</code>.
```
ddorwin: We only need one "both." It doesn't read well this way.

index.html
```diff
@@ -3863,7 +3863,7 @@ <h3 id="h-mediakeystatusmap-interface" resource="#h-mediakeystatusmap-interface"
...
 <span class="idlMethod">    <span class="idlMethType"><span class="hljs-function"><span class="hljs-keyword">void</span></span></span><span class="hljs-function">           </span><span class="idlMethName"><a href="#widl-MediaKeyStatusMap-forEach-void-ForEachCallback-callback"><span class="hljs-function"><span class="hljs-title">forEach</span></span></a></span><span class="hljs-function"> (</span><span class="idlParam"><span class="idlParamType"><a href="#idl-def-ForEachCallback" class="idlType"><code><span class="hljs-function"><span class="hljs-params">ForEachCallback</span></span></code></a></span><span class="hljs-function"><span class="hljs-params"> </span></span><span class="idlParamName"><span class="hljs-function"><span class="hljs-params">callback</span></span></span></span><span class="hljs-function">)</span>;</span>
 };</span></pre>
-      <p>This interface has "entries", "keys", "values" and @@iterator methods brought by <code>iterable</code>.</p>
+      <p>This interface has "entries", "keys", "values", "forEach" and @@iterator methods brought by <code>iterable</code>.</p>
```
ddorwin: Note: This is caused by https://github.com/w3c/respec/pull/804.

encrypted-media-respec.html
jdsmith3000: But it is emphasized...  :)

I'll remove the first "both".

index.html
jdsmith3000: No change or issue here though, correct?

jdsmith3000:
Removed redundant "both" in audio/video capabilities.

index.html
ddorwin: Correct. Just for the record.

PR #205, opened 2016-05-27
Issue #181 - Add explicit monitor CDM algorithm for  CDM-triggered steps

mwatson2:


mwatson2:
We should move the requirement to run this algorithm whilst the object is not closed to `MediaKeySession`.

encrypted-media-respec.html
```diff
@@ -2218,8 +2222,7 @@
 
         <section id="update-key-statuses">
           <h4>Update Key Statuses</h4>
-          <p>The Update Key Statuses algorithm is run when the CDM changes the set of keys <a href="#known-key">known</a> to the session or the status of one or more of the keys.
-          This can happen as the result of a <a def-id="load"></a> or <a def-id="update"></a> call or some other event, such as expiration.
+          <p>The Update Key Statuses algorithm updates the set of keys <a href="#known-key">known</a> to a <a>MediaKeySession</a> or the status of one or more of the keys.
```
ddorwin: nit: "updates the set of keys known to" sounds a bit like this algorithm causes this to change when in reality it is updating the object to reflect the set of known keys. If we think of the `MediaKeySession` as a reflection of CDM state, then maybe that makes more sense.

Changing the wording a bit might help, though. For example, "...`updates the set of <a href="#known-key">known</a> keys for a <a>MediaKeySession</a>...`"

encrypted-media-respec.html
```diff
@@ -2301,9 +2303,92 @@
...
+          <ol>
+            <li>
+              <p>
+                Let <var>session</var> be the <a>MediaKeySession</a> object.
+              </p>
+            </li>
+            <li><p>Let <var>cdm</var> be the CDM instance represented by <var>session</var>'s <var>cdm instance</var> value.</p></li>
+            <li>
+              <p>
+                If <var>cdm</var> has an outgoing message which has not yet been sent, then <a def-id="queue-a-task"></a> to execute the following steps:
```
ddorwin: nit: s/which/that/

encrypted-media-respec.html
```diff
@@ -2301,9 +2303,92 @@
...
+              </p>
+              <ol>
+                <li>
+                  <p>
+                    Let <var>message type</var> and <var>message</var> be the message type and message respectively.
```
ddorwin: nit: comma before "respectively."

encrypted-media-respec.html
```diff
@@ -2301,9 +2303,92 @@
...
-      </section>
 
-      <section>
+        <section id="monitor-cdm">
+          <h4>Monitor for CDM Changes</h4>
+          <p>
+            The Monitor for CDM Changes algorithm executes steps required when various aspects of CDM state change. It is run for each
+            <a>MediaKeySession</a> object in parallel to the main event loop but not in parallel to other procedures defined in this
+            specification that are run in parallel to the main event loop.
+          </p>
```
ddorwin: Should we note that this algorithm only applies to changes that are not covered by other algorithms? IOW, it does not apply to changes that are directly related to an application action. For example, `update()` may result in messages, key status change, and expiration change, but those are all handled within that algorithm.

encrypted-media-respec.html
```diff
@@ -2301,9 +2303,92 @@
             <li><p>Resolve <var>promise</var>.</p></li>
           </ol>
         </section>
-      </section>
 
-      <section>
+        <section id="monitor-cdm">
+          <h4>Monitor for CDM Changes</h4>
```
ddorwin: Should we use "State" in the name? Otherwise, the name could imply just that the CDM changed (i.e. was switched out). "Session State" seems better, but that does not cover the last step. Still, it might be better to be more descriptive and just have that as an extra thing it does.

ddorwin:
LGTM with nits and sugestions.

ddorwin:
nit: Use a colon in the commit message instead of a dash.

jdsmith3000:
Updated "Monitor for CDM State Changes" to address review comments.

ddorwin:
LGTM

Issue #75, opened 2015-07-20
MediaKeyStatusMap: iterable declaration has unexpected behavior; ForEachCallback definition is incorrect


ddorwin:
The spec currently has:
>`callback ForEachCallback = void (BufferSource keyId, MediaKeyStatus status);`

First, values always come before keys in forEach callbacks (modulo what appear to be [bugs](https://www.w3.org/Bugs/Public/show_bug.cgi?id=28978) in Web IDL). Thus, `ForEachCallback`'s parameters should be reversed so the "value" is first and the "key" is second.

<br/>However, I don’t believe either signature is currently possible because `MediaKeyStatusMap` has an `iterable` declaration. Based on my interpretation (see [bug 28978](https://www.w3.org/Bugs/Public/show_bug.cgi?id=28978) and item 2 in [bug 28979](https://www.w3.org/Bugs/Public/show_bug.cgi?id=28979)), the callback would be called with `index, [keyId, status]`. (This may not be the correct ordering - see [bug 28978](https://www.w3.org/Bugs/Public/show_bug.cgi?id=28978).)

In addition, we may not need to explicitly specify `forEach`. See item 4 in [bug 28979](https://www.w3.org/Bugs/Public/show_bug.cgi?id=28979)). Given the complexity, it would probably still be a good idea to explain the format in a non-normative note.

<br/>If our intent is for `MediaKeyStatusMap` to be like a Map, the [iterable declaration](https://heycam.github.io/webidl/#dfn-iterable-declaration) breaks this. We may want to just remove the iterable declaration and specify the "entries", "keys", "values" and `@@iterator` methods explicitly. That would allow us to have a `ForEachCallback` that behaves the same as a `Map`.

<br/>Note: If we continue to use an iterable declaration, we may also want to rework the prose to explicitly identify the “key” and “value” and reference https://heycam.github.io/webidl/#dfn-value-pairs-to-iterate-over like https://fetch.spec.whatwg.org/#headers-class.

ddorwin:
In addition, the iterable declaration results in unexpected behavior for other methods. For example, in the ECMAScript language binding, `keys()` returns an iterator over the indexes (0, 1, 2, ...) and `values()` returns an iterator over the pair of values (`[status, keyId]`).

ddorwin:
[...]

PR #209, opened 2016-05-31
Issue #75: Remove explicit forEach declaration on MediaKeyStatusMap

mwatson2:


ddorwin:
I believe the example needs to be updated such that the "value" comes before the "key." Also, there is an extra `, map` parameter. (This was originally added in https://github.com/w3c/encrypted-media/commit/a7c7cb1d1d7140db2b7bbf61367e2ff8c120d278#diff-f72607e47a6f74e53dc90eab8ee094e2R2407, though I don't recall why.)

I believe the example should be:
> event.target.keyStatuses.forEach(function**(status, keyId)** {

Other than that, LGTM.

jdsmith3000:
That makes sense relative to the Editor's Draft IDL:

> `void forEach(Function callback, optional any thisArg);`

I'll pick up this change to the example.

ddorwin:
LGTM. Thanks. Please change the final commit message to use `Fix #75: ...`

Issue #176, opened 2016-04-28
Add event handler attributes (e.g. onkeystatuseschange on MediaKeySession)


foolip:
Justification is the same as in https://github.com/w3c/media-source/issues/66 and this issue should probably block on the resolution of that.

ddorwin:
Based on https://github.com/w3ctag/design-principles/issues/29#issuecomment-215560292, we should have event handler attributes for all of our events.

We need to add attributes for both [`MediaKeySession` events](https://w3c.github.io/encrypted-media/#mediakeysession-events):
* `onkeystatuseschange`
* `onmessage`

Of the [`HTMLMediaElement` events](https://w3c.github.io/encrypted-media/#htmlmediaelement-events):
* We already have  `onencrypted`.
* We need to add `onwaitingforkey`.

jdsmith3000:
This is a clearly defined change that is ready for implementation.

foolip:
See https://github.com/w3c/html/issues/335#issuecomment-222329752 for a detail you'll need to decide on.

ddorwin:
@foolip, does this only apply to the two `HTMLMediaElement` event handlers?

@domenic said, "So I guess if it bubbles it belongs there." Related to bubbling, the spec text currently says:
* "Queue a task to create an event named encrypted that **does not bubble*** and is not cancellable..."
* "Queue a task to **fire a simple event** named waitingforkey at the media element."
  - Per HTML5, "Firing a simple event named e means that a trusted event with the name e, which does not bubble (except where otherwise stated) and is not cancelable (except where otherwise stated)..."

Thus, it would seem there is no bubbling and thus no need to put these on the `GlobalEventHandlers`. However, HTML5 also uses the "fire a simple event" language for `play`, `playing`, and likely others, which seems inconsistent with @domenic's guidance.

foolip:
Yeah, all events in HTML are on `GlobalEventHandlers` even if they don't bubble, for simplicity I suppose.

I would go with extending `HTMLMediaElement` only, since it's easier to move in the direction of `GlobalEventHandlers` than moving away from it.

ddorwin:
@jdsmith3000, this is ready for implementation per the original plan.

jdsmith3000:
The original plan included adding attributes for MediaKeySession  events:
• onkeystatuseschange 
• onmessage 

And HTMLMediaElement  events:
•Done already: onencrypted .
•Need: onwaitingforkey .

[...]

PR #215, opened 2016-06-02
Issue #176: Add eventHandler attributes

jdsmith3000:


encrypted-media-respec.html
```diff
@@ -2547,7 +2557,12 @@
 
         <dt>attribute EventHandler onencrypted</dt>
         <dd>
-          <p>Event handler for the <a def-id="encrypted"></a> event MUST be supported by all HTMLMediaElements as both a content attribute and an IDL attribute.</p>
```
ddorwin: Is there a reason you removed the "both" part? This was originally added in https://dvcs.w3.org/hg/html-media/rev/a307fdd7e7bd. @adrianba, do you remember why we needed this?

Related, @cpearce filed https://www.w3.org/Bugs/Public/show_bug.cgi?id=25911 to make sure an IDL attribute reflected the corresponding content attribute. We addressed that specific attribute, but that item was later removed from the spec.

@foolip, do you know if both these items are implied for HTML elements? https://www.w3.org/TR/html5/webappapis.html#event-handlers could be read that way?

ddorwin: See also https://www.w3.org/Bugs/Public/show_bug.cgi?id=23897 where the IDL does not imply markup support.

Note: This discussion only applies to the `HTMLMediaElement` handlers.

ddorwin:
Looks good, but I have questions about what, if anything, we need to say about the content attributes on `HTMLMediaElement`.

encrypted-media-respec.html
foolip: It's not implied, I think that https://html.spec.whatwg.org/#event-handlers-on-elements,-document-objects,-and-window-objects is the important bit, normatively. I guess your job might actually be easier if you go back to `GlobalEventHandlers` because then you can just copy that wording.

ddorwin: Thanks. The simplest thing to do is probably to keep this text^ and duplicate it for the new handler. 

^ We should fix it up to add a break between "event" and "MUST". For example, just add "It MUST... to the new text."

jdsmith3000:
I discussed this with @adrianba. He said that the "both a content attribute" phrase requires the event handler be supported for use in HTML, as well as for JS through the IDL.  That suggests we should restore it for `onencrypted` and consider adding it for `waitingforkey`. 

> Event handler for the encrypted event MUST be supported by all HTMLMediaElements as both a content attribute and an IDL attribute.

ddorwin:
Agreed. I don't think the grammar was correct before, so I propose:
>`<p>Event handler for the <a def-id="encrypted"></a> event. It MUST be supported by all HTMLMediaElements as both a content attribute and an IDL attribute.</p>`

For consistency with `onencrypted` and all existing HTMLMediaElements, I believe we should also allow `onwaitingforkey` to be a content attribute, meaning we should add the second sentence above there as well.

jdsmith3000:
@ddorwin Restored with your adjusted language to `onencrypted` and added to `onwaitingforkey`.

encrypted-media-respec.html
```diff
@@ -2557,12 +2557,12 @@
...
         <dt>attribute EventHandler onencrypted</dt>
         <dd>
-          <p>Event handler for the <a def-id="encrypted"></a> event.</p>
+          <p>Event handler for the <a def-id="encrypted"></a> event. It MUST be supported by all HTMLMediaElements as both a content attribute and an IDL attribute.</p> 
         </dd>
 
         <dt>attribute EventHandler onwaitingforkey</dt>
         <dd>
-          <p>Event handler for the <a def-id="waitingforkey"></a> event.</p>
+          <p>Event handler for the <a def-id="encrypted"></a> event. It MUST be supported by all HTMLMediaElements as both a content attribute and an IDL attribute.</p>
```
ddorwin: Fix the def-id: `waitingforkey`

encrypted-media-respec.html
```diff
@@ -2557,12 +2557,12 @@
 
         <dt>attribute EventHandler onencrypted</dt>
         <dd>
-          <p>Event handler for the <a def-id="encrypted"></a> event.</p>
+          <p>Event handler for the <a def-id="encrypted"></a> event. It MUST be supported by all HTMLMediaElements as both a content attribute and an IDL attribute.</p>
```
ddorwin: Extra space at the end of the line. Same below.

ddorwin:
There is a copy-and-paste error. Other than that, LGTM with nit(x2).

Issue #168, opened 2016-04-16
Privacy: Ensure the user can clear all identifiers and other data that might enable tracking


ddorwin:
The spec requires that _Distinctive_ Identifiers be clearable but not _all_ potential identifiers. Examples include:
* Identifiers that don't match the definition of Distinctive Identifier (i.e. random values, especially after #117 is fixed).
* Persisted sessions and session data (#session-storage). (See [bug 27268 comment 5](https://www.w3.org/Bugs/Public/show_bug.cgi?id=27268#c5) by @hsivonen.)

<br/>In addition, as discussed in [bug 27268 comment 6](https://www.w3.org/Bugs/Public/show_bug.cgi?id=27268#c6), "we should add a note somewhere explaining [how] persistent sessions could be used to track users." To address this, we should explicitly mention persistent sessions and other persistent state in https://w3c.github.io/encrypted-media/#user-tracking is the most likely candidate.

It probably also makes sense to explicitly mention persistent sessions in:
* https://w3c.github.io/encrypted-media/#privacy-storedinfo
* https://w3c.github.io/encrypted-media/#incomplete-clearing
* Maybe https://w3c.github.io/encrypted-media/#private-browsing.

ddorwin:
In addition, all traces of a session should be removed when the session data is removed (e.g. a call to `update()` containing the acknowledgement of a release message).

ddorwin:
https://w3c.github.io/encrypted-media/#allow-identifiers-cleared says:
> Implementations should allow users to clear Distinctive Identifier(s) along with cookies [COOKIES] and other site data. It is recommended that users be offered the option to clear Distinctive Identifier(s) as part of user agent features to clear browsing history.

This same guidance should apply to all identifiers and persisted data. Specifically, similar text should be included in or referenced from:
* https://w3c.github.io/encrypted-media/#user-tracking
* https://w3c.github.io/encrypted-media/#privacy-storedinfo
* https://w3c.github.io/encrypted-media/#incomplete-clearing

Note: Addressing this may involve moving it from its existing location and referencing it from there.

mwatson2:
https://w3c.github.io/encrypted-media/#privacy-storedinfo is already quite general. I'm not sure if needs modification for this issue.

Equally, https://w3c.github.io/encrypted-media/#incomplete-clearing consistently refers to "Distinctive Identifiers _and data stored by Key Systems_".

mwatson2:
I created pull request #173 for this.

[...]

PR #216, opened 2016-06-03
Issue #168: Emphasize that all data associated with a session must be cleared in update()

ddorwin:


encrypted-media-respec.html
```diff
@@ -2487,6 +2487,9 @@
           Specifically, the CDM SHOULD NOT store session data during the <a def-id="generateRequest"></a> algorithm.
           This ensures that the application is aware of the session and knows it needs to eventually remove it.
         </p>
+        <p><em>All</em> data associated with a session MUST be cleared when the session is cleared, such as when a <a def-id="remove"></a> request is completed in the <a def-id="update"></a> algorithm. 
+          See <a href="#persistent-state-requirements">Persistent Data</a>.
+        </p>
```
mwatson2: I'm not sure that we should consider the processing of the acknowledgement in the `update()` as 'completion' of the `remove()`. `remove()` "removes all license / keys" and this is completed during the execution of the `remove()` itself.

Perhaps:

> **All** data associated with a session MUST be cleared when the session is closed in `update` due to the processing of an acknowledgement of the _record of key usage_ or _record of license destruction_.

paulbrucecotton:
@ddorwin - Please assign a Milestone whenever new late issues like this are created.

mwatson2:
@paulbrucecotton Actually, this is a Pull Request associated with issue #168 which is V1,

paulbrucecotton:
My bad.  Not enough early morning coffee.

encrypted-media-respec.html
ddorwin: I'm fine with that approach. I want to avoid limiting this language to the current session types and algorithms to avoid rot. Thus, I would like to keep a general statement and example. I've made minor modifications to your suggestion below.

> *All* data associated with a session MUST be cleared when the persisted session is cleared, such as in `update` when the processing of an acknowledgement of the _record of key usage_ or _record of license destruction_.

jdsmith3000: LGTM

PR #217, opened 2016-06-03
Issue #181: Some algorithms are always run inside a task tasks

ddorwin:
This ensures they always run on the main event loop, especially when initiated
from "in parallel" steps. It also ensures ths ordering of any affected promises
is consistent.

The affected algorithms are:
 * Update Key Statuses
 * Update Expiration
 * Session Closed

### DIFF
--- a/encrypted-media-respec.html
+++ b/encrypted-media-respec.html
@@ -1411,7 +1411,7 @@
           <p>The following step is run:</p>
           <ol>
             <li>
-              <p>For each <a>MediaKeySession</a> created by the <var>media keys</var> on which the <a def-id="session-closed-algorithm"></a> algorithm has not been run, run the <a def-id="session-closed-algorithm"></a> algorithm.</p>
+              <p>For each <a>MediaKeySession</a> created by the <var>media keys</var> on which the <a def-id="session-closed-algorithm"></a> algorithm has not been run, <a def-id="queue-a-task"></a> to run the <a def-id="session-closed-algorithm"></a> algorithm.</p>
             </li>
           </ol>
         </section>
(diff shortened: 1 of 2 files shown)


mwatson2:
It's not necessary to queue a task to run an algorithm if we are already inside a task (in particular, if we are already inside one of the MediaKeySession algorithms and we want to call another one.)

In fact, it is important not to do this, so we can be sure attributes are updated before the promise handlers are run.

encrypted-media-respec.html
```diff
@@ -1732,7 +1732,7 @@
                         If the loaded session contains information about any keys (there are <a href="#known-key">known keys</a>), run the <a def-id="update-key-statuses-algorithm"></a> algorithm on the <var>session</var>, providing each key's <a def-id="key-id"></a> along with the appropriate <a>MediaKeyStatus</a>.
                       </p>
                       <p>Should additional processing be necessary to determine with certainty the status of a key, use <a def-id="status-status-pending"></a>.
-                        Once the additional processing for one or more keys has completed, run the <a def-id="update-key-statuses-algorithm"></a> algorithm again with the actual status(es).
+                        Once the additional processing for one or more keys has completed, <a def-id="queue-a-task"></a> to run the <a def-id="update-key-statuses-algorithm"></a> algorithm again with the actual status(es).
```
mwatson2: We are already in a task here (line 1725), so we do not need to queue a task and in fact doing so would result in attribute updates happening after promise handlers.

encrypted-media-respec.html
```diff
@@ -1887,7 +1887,7 @@
                               </p>
                               <p>
                                 Should additional processing be necessary to determine with certainty the status of a key, use <a def-id="status-status-pending"></a>.
-                                Once the additional processing for one or more keys has completed, run the <a def-id="update-key-statuses-algorithm"></a> algorithm again with the actual status(es).
+                                Once the additional processing for one or more keys has completed, <a def-id="queue-a-task"></a> to run the <a def-id="update-key-statuses-algorithm"></a> algorithm again with the actual status(es).
```
mwatson2: Already in a task (1872), so same comment as line 1735.

encrypted-media-respec.html
```diff
@@ -2401,7 +2404,7 @@
                 </li>
                 <li>
                   <p>
-                    Run the <a def-id="update-key-statuses-algorithm"></a> algorithm, passing <var>session</var> and <var>statuses</var>.
+                    <a def-id="Queue-a-task"></a> to run the <a def-id="update-key-statuses-algorithm"></a> algorithm, passing <var>session</var> and <var>statuses</var>.
```
mwatson2: Already in a task (line 2395), so this is unnecessary, though there are no negative side-effects of doing so.

encrypted-media-respec.html
```diff
@@ -2418,7 +2421,7 @@
                 </li>
                 <li>
                   <p>
-                    Run the <a def-id="update-expiration-algorithm"></a> algorithm, passing <var>session</var> and <var>expiration</var>.
+                    <a def-id="Queue-a-task"></a> to run the <a def-id="update-expiration-algorithm"></a> algorithm, passing <var>session</var> and <var>expiration</var>.
```
mwatson2: Already in a task (line 2414), so ditto line 2487.

ddorwin:
Thanks for catching those. It was not my intent to queue tasks when not necessary. I have reverted all four.

mwatson2:
Looks good to me.

PR #229, opened 2016-06-07
Fix #226: Restore reference to "Allow Persistent Data to Be Cleared" in "Get Consent Status"

### LINKED ISSUE #226: Normatively referenced recommendations in Allow Identifiers to Be Cleared algorithm have been moved
The behavior of the Get Consent Status algorithm depends on whether the "any of the recommendations of Allow Identifiers to be Cleared are not supported." These recommendations were moved to a new section in https://github.com/w3c/encrypted-media/commit/5f51c491baa1ebf4360550d574ffd5484f1d95d3.

We need to do something to restore the intended unambiguous reference to these recommendations. Options include adding a reference to the new Allow Persistent Data to Be Cleared section at the same point or somehow noting that the already-referenced section includes everything in the new section.

To help avoid such issues in the future, we should also add a comment to the source that notes that the section is referenced by an algorithm.

jdsmith3000:


ddorwin:
LGTM. Thanks.

Issue #117, opened 2015-10-29
Refine the definition of Distinctive Identifier


ddorwin:
This issue tracks finalization of the [definition of Distinctive Identifier](https://w3c.github.io/encrypted-media/#distinctive-identifier), including addressing the remaining issues from [Bug 27268](https://www.w3.org/Bugs/Public/show_bug.cgi?id=27268).

**April 15, 2016 Update:**
* https://github.com/w3c/encrypted-media/issues/117#issuecomment-210692963 summarizes the problems with the current definition of Distinctive Identifier and explores what it should address.
* https://github.com/w3c/encrypted-media/issues/117#issuecomment-210698890 covers the remaining issues from [Bug 27268](https://www.w3.org/Bugs/Public/show_bug.cgi?id=27268).

steelejoe:
@ddorwin do you have a first cut of the list of issues you can put here for discussion? This is likely to be a long/deep conversation and I would like to help.

ddorwin:
Hi @steelejoe. I have started work on this, but it's not quite there yet.

Since this topic is already complex, please file a separate issue for discussion of that sentence from 8.1.2.

ddorwin:
In addition to the remaining issues from the original bug (see the next comment), the [current definition of Distinctive Identifier](https://cdn.rawgit.com/w3c/encrypted-media/50ebd0cdd9b3f60529707a6f2560d748e74c443e/index.html#distinctive-identifier) is overly broad. For example, it treats a truly random value, most persistent state, and even a cookie the same as a value based on a permanent hardware identifier (as [discussed in the original bug](https://www.w3.org/Bugs/Public/show_bug.cgi?id=27268#c6)). This is inconsistent with the intent, waters down the meaning, and reduces incentives and advantages of avoiding the latter. We need to correct this while still providing the user a level of privacy that is as consistent as practical with the rest of the web platform. The following is a summary of my current thinking. I am working on how to address this in the spec text.


<br/>There are somewhat conflicting goals: Require identifiers used by the Key System implementation to be like cookies and other web storage (unique per origin, clearable, non-associable, etc.) while still allowing "reasonable" functionality, such as attestation, as long as the user consents.

As with all identifiers, I believe the concern is misuse, especially tracking. While any reasonably stable identifier will allow a site to track the user until he or she clears it (cookies already support this), what is more concerning is:
* The ability to track across sites (origins).
[...]

PR #230, opened 2016-06-08
Issue #117: Clarify the meaning of "associable" for Distinctive Identifiers

ddorwin:
Most Distinctive Identifiers are by definition associable by the server that
generated them. This commit attempts to clarify this exception while maintaining
the non-associability requirements for all other entities.


ddorwin:
This was tricky, so feedback welcome. We almost need a separate pair of "associable" and "non-associable" terms and definitions.

jdsmith3000:
@ddorwin I've been working on reviewing this.  It's deep enough that I want to confirm with some others.  I will update Wednesday morning.

@steeljoe Was your specific concern about using a device secret to build a distinctive identifier, and the requirements that would apply to that?

steelejoe:
@jdsmith3000 -- Basically yes. I wanted to confirm that what we implemented for per-origin Distinctive Identifiers could comply with all of the privacy requirements. I believe the new text is clear about what is not allowed and still allows for the performance/robustness gains you can get with individualization. I am happy with this.

encrypted-media-respec.html
```diff
@@ -344,14 +342,22 @@
               <li><p>XORing the origin with a fixed value (because it is trivially reversible)</p></li>
             </ul>
           </div>
+          <p>
+            <a def-id="distinctive-identifiers"></a> may be considered <a def-id="non-associable"></a> as long as such correlation or association is only possible by the entity, such as an <a href="#privacy-individualization">individualization</a> server, that originally generated the <a def-id="distinctive-identifiers"></a> value.
```
mwatson2: This is a little confusing because the definition of "associable" is absolute: something is either associable or not. Distinctive Identifiers certainly meet the definition of "associable" but here we say they may be considered "non-associable".

Perhaps we need to refine the definition of "associable" to be "associable by an entity" (i.e. "two identifiers are associable by an entity if that entity can ..."). Then we will have:
- An identifier is 'associable' if it is associable by any entity.
- Distinctive Identifiers are associable identifiers that are not associable by the application

ddorwin: Thanks for the suggestion. I will try to implement this.

jdsmith3000: This makes sense to me as well.

ddorwin:
I have updated this with @mwatson2's recommendation. Please review so we can move forward with the remaining items in #117.

ddorwin:
If there are no other comments, I will merge this tomorrow morning PDT.

PR #232, opened 2016-06-08
Fix #225: Use the CDM to store record of key usage and clarify asynchronicity

### LINKED ISSUE #225: Access of session's "record of key usage" in the MediaKeySession Destroyed and Session Closed algorithms is inconsistent
The current access of a session's _record of key usage_ are:

1. `generateRequest()`, `remove()`, and `update()` algorithms: By the CDM and in parallel to the main event loop
1. Attempt to Decrypt algorithm: By the CDM within the playback algorithms, which must run in parallel to the main event loop
1. Session Closed algorithm: By the **user agent** in a task, which runs **on the main event loop**
1. MediaKeySession Destroyed algorithm: By the **user agent** in parallel to the main event loop

This session state is managed by the CDM, which always operates in parallel to the main event loop. Thus, it should either be explicitly provided to the user agent algorithm by the CDM (i.e. "Run the Foo Bar algorithm, passing _record of key usage_.") or accessed by the the CDM in parallel. In the latter case, the CDM could perform the operation or be used to obtain the value.

The fix for (4) could be as simple as moving the last two steps under a "Use the _cdm_ to execute the following steps:" step. (3) is more complex because the algorithm is running in a task on the main event loop. If ordering with the other events and the promise in that method are important, just running the step in parallel could be problematic.

mwatson2:


jdsmith3000:
LGTM

ddorwin:
Thanks. I will review as soon as I finish work on the remaining V1 issues.

encrypted-media-respec.html
```diff
@@ -2493,13 +2493,25 @@
             <li><p>Let <var>session</var> be the associated <a>MediaKeySession</a> object.</p></li>
             <li>
               <p>
-                If <var>session</var>'s <var>session type</var> is <a def-id="persistent-usage-record-session"></a>,
-                store <var>session</var>'s <var>record of key usage</var>, if it exists.
+                If <var>session</var>'s <var>session type</var> is <a def-id="persistent-usage-record-session"></a>, execute the following steps:
```
ddorwin: "...  in parallel"

This is the only remaining CDM use that is not in parallel. The rest of this algorithm is running (in a task) on the main event loop.

mwatson2: As per [this comment](https://github.com/w3c/encrypted-media/issues/225#issuecomment-224280044), it makes no difference: "in parallel" means only that page-visible effects don't happen immediately, but there are no page-visible effects. So, instead, we have the note below.

PR #234, opened 2016-06-08
Issue #110: Allow Distinctive Identifiers for Per-Origin Individualization

### LINKED ISSUE #110: Individualization text regarding device identifiers is overly broad and restrictive
Section 11.4.3.2 contains the following text:
"Such implementations MUST NOT use Distinctive Identifier(s) for a device or user of a device in the individualization process."

This is too broad. I proposed instead the following:
"Such implementations MUST NOT directly provide Distinctive Identifier(s) for a device or user of a device in any messages sent during the individualization process. "

This allows for implementations which generate unique identifiers not directly associable with the device or user by digesting a mixture of device identifiers. These identifiers can have the security property that two different devices are unlikely to generate the same identifier, but also have the privacy property that it is very difficult to match an identifier to a user+device.

David suggested looking for guidance to [this email thread](https://lists.w3.org/Archives/Public/public-html-media/2014Oct/0092.html).

Original bug was [Bug 27168](https://www.w3.org/Bugs/Public/show_bug.cgi?id=27168).

(updated to reflect comment 1)

ddorwin:


ddorwin:
If there are no comments, I will likely merge this soon to unblock other PRs. Please let me know if you are planning to review.

jdsmith3000:
Reviewing now.

encrypted-media-respec.html
```diff
@@ -3825,25 +3825,25 @@
 
           <section id="per-origin-individualization">
             <h5>Per-Origin Individualization</h5>
-            <p>Per-<a def-id="origin"></a> individualization is performed between the <a def-id="cdm"></a> and an application-selected server and results in a per-origin identifier.
-              The process MUST be performed via these APIs and MUST NOT involve other types of communication.
-              Any <a def-id="distinctive-identifiers"></a> used MUST be origin-specific.
+            <p>Per-<a def-id="origin"></a> individualization is performed between the <a def-id="cdm"></a> and the application, including an application-selected server, and results in a per-origin identifier.
+              The process MUST be performed via these APIs and MUST NOT involve other methods of communication.
+              As with all other uses of the APIs, <a def-id="distinctive-identifiers"></a> MAY be used but indelible or non-origin-specific MUST NOT be exposed, even in encrypted form.
```
mwatson2: s/indelible or non-origin-specific/indelible or non-origin-specific identifiers/

encrypted-media-respec.html
```diff
@@ -3825,25 +3825,25 @@
...
+              If one or more <a def-id="distinctive-identifiers"></a> is used, the resulting identifier is by definition also a <a def-id="distinctive-identifier"></a>.
```
mwatson2: s/is/are/ ?

mwatson2: Is this strictly true ? The approach of hashing one identifier with a smaller random salt can create a new identifier from a space that is large enough that it is not associable but small enough that it has useful non-migratability properties.

encrypted-media-respec.html
```diff
@@ -3825,25 +3825,25 @@
...
+              <li><p>MUST be passed to or through the application via these APIs.</p></li>
+              <li><p>SHALL use the message type <a def-id="message-type-individualization-request"></a> for all related <a def-id="message"></a> events.</p></li>
               <li><p>MUST NOT be performed by the user agent.</p></li>
               <li><p>MUST NOT be performed directly by the CDM.</p></li>
-              <li><p>MUST NOT contain non-origin-specific per-client information, such as a <a def-id="distinctive-identifier"></a>.</p></li>
-              <li><p>As with all other uses of the APIs, responses passed to the CDM MUST NOT contain executable code.</p></li>
+              <li><p>MUST NOT contain non-origin-specific per-client information or any type of indelible or semi-permanent identifer.</p></li>
+              <li><p>MUST adhere to the <a href="#identifier-requirements">identifier requirements</a>, including exposing only values that are <a href="#per-origin-identifiers">unique per origin</a> and <a href="#allow-identifiers-cleared">clearable</a> and <a href="#encrypt-identifiers">encrypting</a> <a def-id="distinctive-identifiers"></a>.</p></li>
```
mwatson2: Do we really need the "including ..." subclause ? This duplicates the requirements which is not really a good idea. They are clear in the referenced section.

encrypted-media-respec.html
jdsmith3000: It does seem possible here for the resulting identifier to be not distinctive.

jdsmith3000:
This section LGTM in general, assuming the comment about producing a non-distinctive identifier from one or more distinctive ones.

encrypted-media-respec.html
ddorwin: Yes, but I will use "values."

encrypted-media-respec.html
ddorwin: I believe this follows from the following bullet in the definition of Distinctive Identifier:
>The generation, individualization, provisioning or other process that produced the value involved, used, provided, derived from, or similarly involved such an indelible value or **another distinctive identifier.**

I'm open to considering scenarios like @mwatson2 suggests, but I think we should evaluate this separately. Shall we file a V1NonBlocking issue, reference it in an ISSUE box in the definition, and move forward on this CL, which is almost entirely unrelated to that issue?

@mwatson2, if you agree, please file the issue.

encrypted-media-respec.html
ddorwin: Done. I commented it out within a TODO to address when the referenced section is refactored per #117.

ddorwin:
I have a proposal for dealing with the "by definition" question. Please review and reply. All other issues are fixed.

mwatson2:
@ddorwin I don't see your proposal on the "by definition" question. Or did you mean you're working on one ?

ddorwin:
@mwatson2, it is hidden by default by the GitHub UI. See https://github.com/w3c/encrypted-media/pull/234#discussion_r66361853.

ddorwin:
Per the discussion at https://github.com/w3c/encrypted-media/issues/110#issuecomment-224774599, I will merge this and we will continue work on #110.

PR #235, opened 2016-06-09
Fix #231: Ensure both individualization types produce compliant Distinctive Identifiers

### LINKED ISSUE #231: The Origin-Independent Individualization process would not generate a compliant Distinctive Identifier
The current Origin-Independent Individualization process currently says the following:

1. "The resulting identifier MUST be origin- and application-independent."
1. "Implementations MAY derive non-associable per-origin identifiers from such identifiers and provide those to the application (encrypted)."

I think this text is outdated and based on an old definition of Distinctive Identifier before privacy-related requirements were added. As with "Per-Origin Individualization" in #110, I think this section needs to be updated.

Distinctive Identifiers must be per-origin, so (1) means the value cannot be used directly. (Per the parent section, this process is supposed to provide a Distinctive Identifier, which is not the case.) However, I believe any derivation as mentioned in (2) would either a) be associable (e.g. because the interim value is used to sign the derived identifiers and maintain the attestation chain or b) break the usefulness of such an identifier for attestation.

I believe what is really intended is "Direct Individualization" where we allow indelible identifiers to be used because they do not go through the application. Similarly, "Per-Origin Individualization" should be "App-Based Individualization."

I will make an attempt at this and create a PR for review.

ddorwin:
Also renames the two types to more accurately reflect their behavior.

### DIFF
--- a/encrypted-media-respec.html
+++ b/encrypted-media-respec.html
@@ -174,7 +174,7 @@
 
           <p>All messages and communication to and from the CDM, such as between the CDM and a license server, MUST be passed through the user agent.
             The CDM MUST NOT make direct out-of band network requests.
-            All messages and communication other than those described in <a href="#origin-independent-individualization">Origin-Independent Individualization</a> MUST be passed through the application via the APIs defined in this specification.
+            All messages and communication other than those described in <a href="#direct-individualization">Direct Individualization</a> MUST be passed through the application via the APIs defined in this specification.
             Specifically, all communication that contains application-, <a def-id="origin"></a>-, or content-specific information or is sent to a URL specified by the application or based on its origin, MUST pass through the APIs.
             This includes all license exchange messages.
           </p>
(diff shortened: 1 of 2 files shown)


encrypted-media-respec.html
```diff
@@ -3841,11 +3835,14 @@
               <li><p>MUST NOT be passed to or through the application via these APIs.</p></li>
               <li><p>MUST be sent to a URL selected independently of any origin and application.</p></li>
             </ul>
+            <p>
+              Implementations MUST NOT expose, even in encrypted form, <a def-id="origin">origin(s)</a>-, origin- or application-specific information, or values that are <a def-id="associable"></a> with orign(s) to centralized servers since this could create a central record of all origins visited by a user or device.
```
mwatson2: s/orign(s)/origin(s)/

encrypted-media-respec.html
```diff
@@ -3860,7 +3857,11 @@
...
               <li><p>MUST NOT provide executable code to the CDM.</p></li>
             </ul>
-            <p>With appropriate precautions, such individualization can provide better privacy than origin-independent individualization, though not as good as models that do not use <a def-id="distinctive-identifiers"></a>.
+            <p>
+              When <a def-id="associable"></a> values, including <a def-id="distinctive-identifier">Distinctive Identifier(s)</a>, are used in the process,
+              implementations MUST NOT expose, even in encrypted form, <a def-id="origin">origin(s)</a>-, origin- or application-specific information, or values that are <a def-id="associable"></a> with orign(s) to centralized servers since this could create a central record of all origins visited by a user or device.
```
mwatson2: s/orign(s)/origin(s)/

mwatson2:
LGTM with 2 typos.

PR #238, opened 2016-06-09
Issue #117: Define and use Distinctive Permanent Identifier

ddorwin:


encrypted-media-respec.html
```diff
@@ -359,6 +359,34 @@
...
+            A <dfn id="permanent-identifier">Permanent Identifier</dfn> a value, piece of data, implication of the possession of a piece of data, or an observable behavior or timing that is indelible in some way or otherwise non-trivial for the user to remove, reset, or change.
+            This, includes but is not limited to:
+          </p>
+          <ul>
+            <li><p>A hardware or hardware-based identifier</p></li>
+            <li><p>A value provisioned in the hardware device in the factory</p></li>
+            <li><p>A value associated with or derived from the operating system installation instance</p></li>
+            <li><p>A value associated with or derived from the user agent installation instance</p></li>
+            <li><p>A value associated with or derived from the <a def-id="cdm"></a> or other software component</p></li>
+            <li><p>A value in a confguration file or similar semi-permanent data, even if generated on the client</p></li>
```
mwatson2: s/confguration/configuration/

encrypted-media-respec.html
```diff
@@ -456,16 +483,31 @@
...
           </div>
-          <p class="note">The source of the identifier does not affect whether it is distinctive.
-            For example, an identifier that is permanently part of the client device, contained in the CDM, generated on the client, or generated as part of some individualization or other provisioning process is considered distinctive if it meets the criteria above.
+        </dd>
+
+        <dt>Use of Distinctive Identifiers and Distinctive Permanent Identifiers</dt>
+        <dd>
+          <p>
+            An implementation, configuration, instance, or object <dfn id="uses-distinctive-identifiers">uses Distinctive Identifier(s)</dfn> if, at any time during its lifetime or the lifetime of related such entities,
+            it exposes, even in encrypted form, on or more <a def-id="distinctive-identifier">Distinctive Identifier(s)</a>, information about them, or values derived from or otherwise related to them outside the client.
```
mwatson2: s/on/one/

mwatson2: "information about them" is very broad: a boolean indicating whether one was present or not would qualify.

It's not clear to me that any of the text after "Distinctive Identifiers" is needed here. Surely anything which is derived from one Distinctive Identifier that is a cause for concern also meets the definition of Distinctive Identifier ?

encrypted-media-respec.html
```diff
@@ -456,16 +483,31 @@
...
+            This includes but is not limited to providing such a value to the application and/or license, <a href="#privacy-individualization">individualization</a>, or other server.
+          </p>
+          <p>
+            An implementation, configuration, instance, or object <dfn id="uses-distinctive-permanent-identifiers">uses Permanent Distinctive Identifier(s)</dfn> if, at any time during its lifetime or the lifetime of related such entities,
+            it exposes, even in encrypted form, on or more <a def-id="distinctive-permanent-identifier-maybe-plural"></a>, information about them, or values derived from or otherwise related to them outside the client.
+            This includes but is not limited to providing such a value to an <a href="#privacy-individualization">individualization</a> server.
```
mwatson2: Same comment as line 499

encrypted-media-respec.html
```diff
@@ -456,16 +483,31 @@
...
+        <dd>
+          <p>
+            An implementation, configuration, instance, or object <dfn id="uses-distinctive-identifiers">uses Distinctive Identifier(s)</dfn> if, at any time during its lifetime or the lifetime of related such entities,
+            it exposes, even in encrypted form, on or more <a def-id="distinctive-identifier">Distinctive Identifier(s)</a>, information about them, or values derived from or otherwise related to them outside the client.
+            This includes but is not limited to providing such a value to the application and/or license, <a href="#privacy-individualization">individualization</a>, or other server.
+          </p>
+          <p>
+            An implementation, configuration, instance, or object <dfn id="uses-distinctive-permanent-identifiers">uses Permanent Distinctive Identifier(s)</dfn> if, at any time during its lifetime or the lifetime of related such entities,
+            it exposes, even in encrypted form, on or more <a def-id="distinctive-permanent-identifier-maybe-plural"></a>, information about them, or values derived from or otherwise related to them outside the client.
```
mwatson2: Same comment as line 499

encrypted-media-respec.html
```diff
@@ -3172,7 +3213,7 @@
 
         <section id="allow-identifiers-cleared">
           <h3>Allow Identifiers to Be Cleared</h3>
-          <p>As a consequence of the requirements in <a href="#allow-persistent-data-cleared">Allow Persistent Data to Be Cleared</a>, implementations that use <a def-id="distinctive-identifier">Distinctive Identifier(s)</a> MUST allow the user to clear those identifiers such that they are no longer retrievable both outside, such as via the APIs defined in this specification, and on the client device.
+          <p>As a consequence of the requirements in <a href="#allow-persistent-data-cleared">Allow Persistent Data to Be Cleared</a>, implementations that <a href="#uses-distinctive-identifiers-or-distinctive-permanent-identifiers">Distinctive Identifier(s) or Distinctive Permanent Identifier(s)</a> MUST allow the user to clear the <a def-id="distinctive-identifier">Distinctive Identifier(s)</a> and values associated with <a def-id="distinctive-permanent-identifier-maybe-plural"></a> such that they are no longer retrievable both outside, such as via the APIs defined in this specification, and on the client device.
```
mwatson2: I thought the deference between Distinctive Identifiers and Distinctive Permanent Identifiers was that the latter _could not_ be cleared (and hence must not be disclosed to applications).

encrypted-media-respec.html
ddorwin: From above:
>"information about them" is very broad: a boolean indicating whether one was present or not would qualify.

This was adapted from line 416 ("It, information about it, or values derived from or otherwise related to it are exposed, even in encrypted form, outside the client."), which has been been in the Distinctive Identifier definition for a long time. I debated all the parts, but being consistent seemed appropriate.

As for the Boolean, I think it could apply. I don't think we would want an application to be able to ask "Do you have a permanent Foo Key System key on your device?" without triggering some of the behaviors associated with these definitions.

mwatson2:
LGTM with some comments.

encrypted-media-respec.html
ddorwin: See below for more details.

In this case, I think I agree that Distinctive Identifier already includes this and any value derived from a Distinctive Identifier is a Distinctive Identifier, so we can drop this one.

ddorwin: Looking at this again, I think it is mostly correct as is. For example: A Distinctive Identifier is a value for which it or "information about it... [is] exposed." Here, we are defining that such exposure of it or  information about it qualifies as "use."

The one thing that is perhaps redundant is "derived from." We could drop that, but the consistency with the next item is nice.

encrypted-media-respec.html
ddorwin: I don't think this conflicts with that. It may be a little subtle, but this basically adds that:
* If you _use_ Distinctive Permanent Identifiers (but somehow not Distinctive Identifiers), you must allow associated values to be cleared.
* If you _use_ Distinctive Permanent Identifiers, all associated values, not just Distinctive Identifiers, must be clearable.

Let me see if I can separate this to be clearer.

mwatson2: I see what you are trying to say. I think the problem is that I believe Distinctive Permanent Identifiers are Distinctive Identifiers, at least according to the definitions.

This might be a bigger problem than just here, since we place normative requirements on Distinctive Identifiers which Distinctive Permanent Identifiers cannot meet. Or do we place those requirements only on Distinctive Identifiers that are exposed to applications ?

If the latter, you could replace:
> allow the user to clear the Distinctive Identifier(s)
with
> allow the user to clear all Distinctive Identifier(s) that are not Distinctive Permanent Identifier(s)

encrypted-media-respec.html
mwatson2: Ok, I see it is part of the definition. I'll think about this more and raise a VNext issue if I see a problem.

encrypted-media-respec.html
ddorwin: Distinctive Permanent Identifiers must not be Distinctive Identifiers because the latter are clearable, per-origin, and may be exposed to the application. Thus, we need to ensure the definitions don't imply otherwise.

It may also be subtle, but I don't believe the current definitions mean Distinctive Permanent Identifiers are Distinctive Identifiers (even though they are "distinctive" and "identifiers"). The first two sub-bullets in the Distinctive Identifier definition require derivation or generation. Permanent Identifiers are not generated or derived, at least within the scope of this specification. (They are generated in a factory and may be generated or derived as part of, for example, an OS installation, but I don't think that's what we are referring to.)

For now, I have added a NOTE to clarify this, but it relies on that same "scope" argument. Perhaps we can improve the wording to avoid any question.

One option might be to add "Is not a [Distinctive] Permanent Identifier" to the list of ANDed properties of Distinctive Identifier.

jdsmith3000: Does this move away then from the name "Distinctive Permanent Identifiers"?  I think it should.  As it is, to me anyway, it implies a Distinctive Identifier that is Permanent, which poses a number of conflicts with spec requirements.

ddorwin: It certainly isn't a Permanent Distinctive Identifier (note the order). I was originally just using "Permanent Identifier," but, for example, using a permanent per-OS key shouldn't trigger all these behavior. That's why I added "distinctive," but I agree that it could be confusing to someone not familiar with the spec or nuances. Do you have other suggestions?

We can always change the name of terms that don't affect the API later. For now, we just need to define the behavior.

ddorwin:
I just uploaded a new version that attempts to address the discussion. If this looks good, let's merge it and continue to think about ways to add clarity.

encrypted-media-respec.html
```diff
@@ -359,6 +359,37 @@
           </p> 
         </dd>
 
+        <dt>Permanent Identifiers</dt>
+        <dd>
+          <p>
+            A <dfn id="permanent-identifier">Permanent Identifier</dfn> a value, piece of data, implication of the possession of a piece of data, or an observable behavior or timing that is indelible in some way or otherwise non-trivial for the user to remove, reset, or change.
```
mwatson2: s/a value/is a value/

encrypted-media-respec.html
```diff
@@ -370,30 +401,29 @@
...
             </li>
             <li><p>It has one or more the following properties:</p>
               <ul>
-                <li><p>It is derived from a value that is indelible in some way, including but not limited to hardware, operating system installation instance, user agent installation instance, <a def-id="cdm"></a> or other software component, confguration file or similar semi-permanent data, or client or other user account values.</p></li>
-                <li><p>The generation, <a href="#privacy-individualization">individualization</a>, provisioning or other process that produced the value involved, used, provided, derived from, or similarly involved such an indelible value or another distinctive identifier.</p></li>
+                <li><p>It is derived from one or more <a def-id="distinctive-permanent-identifier-maybe-plural"></a>.</p></li>
+                <li><p>The generation, <a href="#privacy-individualization">individualization</a>, provisioning or other process that produced the value involved, used, provided, derived from, or similarly involved one or more <a def-id="distinctive-permanent-identifier-maybe-plural"></a> or another distinctive identifier.</p></li>
```
mwatson2: Pedantically (and not blocking us progressing), an example of a "process that produced a value" could be simply copying. So a copy of a Distinctive Permanent Identifier is a Distinctive Identifier (by the above definition) and since it has the same value as the Distinctive Permanent Identifier the Distinctive Permanent Identifier is also a Distinctive Identifier.

mwatson2: LGTM

encrypted-media-respec.html
jdsmith3000: It's okay to adjust naming later.  I still believe though that it's confusing to apply "Distrinctive" to the Permanent Identifiers name when many of the Distinctive requirements don't apply to it.

PR #239, opened 2016-06-09
Fix #233: Make MediaKeyStatusMap.get return any

### LINKED ISSUE #233: It may not be possible to return undefined from MediaKeyStatusMap.get()
The [definition](https://w3c.github.io/encrypted-media/#widl-MediaKeyStatusMap-get-MediaKeyStatus-BufferSource-keyId) of `MediaKeyStatusMap`'s `get()` method says:
> Returns the `MediaKeyStatus` of the key identified by _keyId_ or `undefined` if the status of the key identified by _keyId_ is not known.

This text was added for #70 with the goal of being consistent with the [ES6 Map interface's get()](https://people.mozilla.org/~jorendorff/es6-draft.html#sec-map.prototype.get). We cannot just use `maplike` for the reasons in #25 and #39.

In an implementation discussion, @foolip said, "The IDL says that the return value is a MediaKeyStatus, and undefined is not of that type."

Is there a way we can be consistent with the behavior of `Map`?

mwatson2:


ddorwin:
LGTM to reflect what we want, though I'm not sure this is allowed by the Web IDL grammer. We may need to reopen and readdress #233 depending on the outcome of that discussion.

ddorwin:
Not LGTM. See https://github.com/w3c/encrypted-media/issues/233#issuecomment-225016042.

ddorwin:
LGTM

ddorwin:
nit: Use `()` in the commit message: `MediaKeyStatusMap.get()`

PR #240, opened 2016-06-09
Individualization: Use the new "use" and Distinctive Permanent Identifier definitions

ddorwin:
Just some minor cleanup to use the terms from PR #238, addressing the TODO.


mwatson2:
LGTM

PR #241, opened 2016-06-09
Issue #117: Update identifier requirements to cover all types of identifiers

ddorwin:
This should be the last (major) change for #117.

### DIFF
--- a/encrypted-media-respec.html
+++ b/encrypted-media-respec.html
@@ -3177,7 +3177,12 @@
             </li>
             <li><p><a def-id="distinctive-identifiers"></a> MUST be <a href="#encrypt-identifiers">encrypted</a> when exposed outside the client, <a href="#per-origin-identifiers">unique per-origin</a>, and <a href="#allow-identifiers-cleared">clearable</a>.</p></li>
             <li><p><a def-id="distinctive-permanent-identifiers"></a> MUST be <a href="#encrypt-identifiers">encrypted</a> when exposed outside the client and MUST NOT be exposed to the application.</p></li>
-            <li><p>All potential identifiers or distinctive values not covered above that are generated as a result of use of these APIs (e.g. persisted session data) MUST be <a href="#per-origin-identifiers">unique per-origin</a> and <a href="#allow-identifiers-cleared">clearable</a>.</p></li>
+            <li>
+              <p>
+                All potential identifiers or distinctive values not covered above that are generated as a result of use of these APIs MUST be <a href="#per-origin-identifiers">unique per-origin</a> and <a href="#allow-identifiers-cleared">clearable</a>.
+                This includes but is not limited to random identifiers, session data, and other CDM data.
+              </p>
+            </li>
           </ul>
         </div>
 
@@ -3195,12 +3200,14 @@
             <li>
               <p>
                 Implementations that <a href="#uses-distinctive-identifiers-or-distinctive-permanent-identifiers">use Distinctive Identifier(s) or Distinctive Permanent Identifier(s)</a> SHOULD support the option to not use them.
+                <!-- TODO(#243): Decide whether to include this statement.
                 Implementations with such support SHOULD expose the ability for the user to select this option.
+                -->
               </p>
               <div class="note">
                 <p>
                   When supported, applications can select for this mode using <a def-id="option-distinctiveIdentifier"></a> = <a def-id="requirement-not-allowed"></a>.
-                  Selecting such an option MAY affect the results of the <a def-id="requestMediaKeySystemAccess"></a> call and/or the license requests that are generated from subsequently generated sessions.
+                  Selecting such an option may affect the results of the <a def-id="requestMediaKeySystemAccess"></a> call and/or the license requests that are generated from subsequently generated sessions.
                 </p>
                 <p>
                   Providing the user access to select or choose this implementation capability may allow the user to access content while maintaining a higher degree of privacy.
@@ -3218,12 +3225,14 @@
             The encryption MUST ensure that the ciphertext cannot be used as a proxy for the actual identifier, even given the same plaintext.
           </p>
           <p class="issue"><a href="https://github.com/w3c/encrypted-media/issues/219">Issue 219</a> - Add more specific text about ensuring the desired privacy properties when encrypting identifiers.</p>
-          <p class="note">Identifiers may be exposed in the following ways:</p>
+          <div class="note">
+            <p>Identifiers may be exposed in the following ways:</p>
             <ul>
               <li><p>To the application via a <a def-id="message"></a> event.</p></li>
               <li><p>In a message from a server, such as one that is passed to <a def-id="update"></a>.</p></li>
               <li><p>As part of <a href="#privacy-individualization">individualization</a>.</p></li>
             </ul>
+          </div>
           <p>
             The <a def-id="cdm"></a> MUST verify that the encryption key belongs to a valid server for its Key System.
             For identifers exposed to the application, this MAY be implemented using a <a href="#server-certificate">server certificate</a>.
@@ -3244,7 +3253,8 @@
 
         <section id="per-origin-identifiers">
           <h3>Use Per-Origin Identifiers</h3>
-          <p>          
+          <p class="issue"><a href="https://github.com/w3c/encrypted-media/issues/242">Issue 242</a> - Add a section that covers this topic for values and data that are not identifiers.</p>
+          <p>
             All potential identifiers or distinctive values that are not <a href="#permanent-identifier">Permanent Identifiers</a> MUST be unique per <a def-id="origin"></a>.
             That is, the identifier(s) used for one <a def-id="origin"></a> using these APIs MUST be different from those used for any other origin using the APIs.
           </p>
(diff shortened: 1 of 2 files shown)


encrypted-media-respec.html
```diff
@@ -3165,7 +3165,21 @@
...
+          <p>In summary:</p>
+          <ul>
+            <li><p><a href="#limit-or-avoid-use-of-distinctive-identifiers-and-permanent-identifiers">Limit or Avoid use of Distinctive Identifiers and Permanent Identifiers</a>.</p></li>
+            <li>
+              <p>
+                All identifers that are not <a href="#permanent-identifier">Permanent Identifiers</a> MUST be <a href="#per-origin-identifiers">unique per-origin</a> and <a href="#allow-identifiers-cleared">clearable</a>.</p>
+                They SHOULD be <a href="#encrypt-identifiers">encrypted</a> when exposed outside the client.
+              </p>
+            </li>
+            <li><p><a def-id="distinctive-identifiers"></a> MUST be <a href="#encrypt-identifiers">encrypted</a> when exposed outside the client, <a href="#per-origin-identifiers">unique per-origin</a>, and <a href="#allow-identifiers-cleared">clearable</a>.</p></li>
```
mwatson2: Aren't the last two requirements duplicative of line 3174 ?

Also, it's strange to have normative looking MUST language in a note (I think notes are by definition non-normative, no?)

encrypted-media-respec.html
```diff
@@ -3179,44 +3193,66 @@
               <p class="note">For example, <a def-id="temporary-session"></a> and <a def-id="persistent-license-session"></a> sessions may have different requirements.</p>
             </li>
             <li>
-              <p>Implementations that <a href="#uses-distinctive-identifiers-or-distinctive-permanent-identifiers">use Distinctive Identifier(s) or Distinctive Permanent Identifier(s)</a> SHOULD support the option to not use them.
-                When supported, applications can select for this mode using <a def-id="option-distinctiveIdentifier"></a> = <a def-id="requirement-not-allowed"></a>.
-                Selecting such an option MAY affect the results of the <a def-id="requestMediaKeySystemAccess"></a> call and/or the license requests that are generated from subsequently generated sessions.
+              <p>
+                Implementations that <a href="#uses-distinctive-identifiers-or-distinctive-permanent-identifiers">use Distinctive Identifier(s) or Distinctive Permanent Identifier(s)</a> SHOULD support the option to not use them.
+                Implementations with such support SHOULD expose the ability for the user to select this option.
```
mwatson2: Is this a new requirement, or moved from somewhere else ? If new, I suggest we discuss it separately.

encrypted-media-respec.html
```diff
@@ -3179,44 +3193,66 @@
...
               </p>
+              <div class="note">
+                <p>
+                  When supported, applications can select for this mode using <a def-id="option-distinctiveIdentifier"></a> = <a def-id="requirement-not-allowed"></a>.
+                  Selecting such an option MAY affect the results of the <a def-id="requestMediaKeySystemAccess"></a> call and/or the license requests that are generated from subsequently generated sessions.
+                </p>
```
mwatson2: The above two lines appear to have moved from normative text into a note. Was this intentional ?

encrypted-media-respec.html
```diff
@@ -3179,44 +3193,66 @@
...
           <p class="issue"><a href="https://github.com/w3c/encrypted-media/issues/219">Issue 219</a> - Add more specific text about ensuring the desired privacy properties when encrypting identifiers.</p>
-          <p>This MAY be implemented using a <a href="#server-certificate">server certificate</a>.</p>
-          <p>The license server MUST NOT expose a <a def-id="distinctive-identifier"></a> to any entity other than the CDM that sent it.</p>
-          <a def-id="distinctive-permanent-identifiers"></a> MUST also be encrypted at the message exchange level.
+          <p class="note">Identifiers may be exposed in the following ways:</p>
+            <ul>
+              <li><p>To the application via a <a def-id="message"></a> event.</p></li>
+              <li><p>In a message from a server, such as one that is passed to <a def-id="update"></a>.</p></li>
+              <li><p>As part of <a href="#privacy-individualization">individualization</a>.</p></li>
+            </ul>
```
mwatson2: I think you intended the list to appear in the note as well, but it [does not](https://rawgit.com/ddorwin/encrypted-media/259b372a13888b2b52f9364bbe0f096fb3d8c4ad/encrypted-media-respec.html#encrypt-identifiers).

encrypted-media-respec.html
```diff
@@ -3228,16 +3264,15 @@
           <h3>Allow Identifiers to Be Cleared</h3>
           <p>
             As a consequence of the requirements in <a href="#allow-persistent-data-cleared">Allow Persistent Data to Be Cleared</a>,
+            all potential identifiers or distinctive values that are not <a href="#permanent-identifier">Permanent Identifiers</a> MUST be clearable
+            such that the values are no longer retrievable, observabe, or inferable both outside, such as via the APIs defined in this specification, and on the client device.
```
mwatson2: s/observabe/observable/

encrypted-media-respec.html
```diff
@@ -3228,16 +3264,15 @@
...
-            <li><p>implementations that <a href="#uses-distinctive-permanent-identifiers">use Distinctive Permanent Identifier(s)</a> and MUST allow the user to clear values associated with the <a def-id="distinctive-permanent-identifier-maybe-plural"></a></p></li>
-          </ul>
           <p>
-            such that the <a def-id="distinctive-identifiers"></a> and other associated values and are no longer retrievable, observabe, or inferable both outside, such as via the APIs defined in this specification, and on the client device.
+            Implementations that <a href="#uses-distinctive-identifiers">use Distinctive Identifier(s)</a> MUST allow the user to clear the <a def-id="distinctive-identifier">Distinctive Identifier(s)</a>.
+            Implementations that <a href="#uses-distinctive-permanent-identifiers">use Distinctive Permanent Identifier(s)</a> and MUST allow the user to clear values associated with the <a def-id="distinctive-permanent-identifier-maybe-plural"></a>.
```
mwatson2: "values associated with" is a bit vague, but I think we could address clarification during CR.

encrypted-media-respec.html
```diff
@@ -3228,16 +3264,15 @@
...
           </p>
           <p>
-            Once cleared, new <a def-id="non-associable-by-application"></a> value(s) MUST be generated when <a def-id="distinctive-identifier">Distinctive Identifier(s)</a> are subsequently needed.
+            Once cleared, new <a def-id="non-associable-by-application"></a> value(s) MUST be generated when values, such as <a def-id="distinctive-identifiers"></a> subsequently needed.
```
mwatson2: s/subsequently/are subsequently/

encrypted-media-respec.html
ddorwin: Yes, there is duplication here. My goal was to have a single place to look and see the requirements for various types of identifiers without having to put a bunch of categories together. That information exists normatively in the sections below - this is intended to help the reader.

jdsmith3000: The last two lines do apply requirements to "distinctive values" and "persisted session data" though.

I don't see a problem with summarizing MUST requirements in note.  It seems implied though that these are not the source of the normative control, but are summarized there for clarity.

ddorwin: I've had the same thought on the RFC words, but we've done this throughout so we should evaluate that overall.

While such notes themselves are non-normative, they are often trying to provide additional information about requirements and I find the indication that "this is a requirement" helpful. This is certainly the most concise way to say "are required to."

encrypted-media-respec.html
ddorwin: I think this specific line is new. That's fine - I'll open a GitHub issue. I'm wondering whether it would be better to include this text and add an ISSUE box that it needs discussion or remove it and assume it is not substantive and could be added. Thoughts?

encrypted-media-respec.html
ddorwin: Yes, this seemed to describe usage and not any additional normative requirements. There is no longer an explicit "you must allow `distinctiveIdentifier` to control this," but I think that is sufficiently covered by that part of the spec. What do you think?

encrypted-media-respec.html
ddorwin: >The last two lines do apply requirements to "distinctive values" and "persisted session data" though.

@jdsmith3000, I don't understand.

encrypted-media-respec.html
ddorwin: Just a note: This text is not new in this PR - it is just moved to paragraph form. Agreed.

encrypted-media-respec.html
mwatson2: Let's just open an issue to track it. I think it could be added later: it's only a recommendation.

encrypted-media-respec.html
mwatson2: Ok, makes sense. Perhaps decapitalize the MAY since it's not a normative permissive statement but just a description of a consequence of algorithms described elsewhere.

ddorwin:
I addressed the comments and added a couple non-normative statements. Please take a look.

mwatson2:
LGTM

PR #254, opened 2016-06-23
Fix #162: Add Conformance Section

### LINKED ISSUE #162: Explicitly state "RFC words" should be interpreted per RFC2119
The spec uses RFC2119 words MUST, MUST NOT, etc., and they are differentiated by ReSpec, but we do not reference RFC2119. We should explicitly document their interpretation like other specs do.

Examples:
* https://www.w3.org/TR/html5/infrastructure.html#conformance-requirements
* https://w3c.github.io/webappsec-secure-contexts/#conventions

As an aside, this seems like something ReSpec should insert automatically along with the formatting.

mwatson2:


ddorwin:
Cool! LGTM

PR #255, opened 2016-06-23
Fix #185: Define 'closed' for MediaKeySession and use throughout

### LINKED ISSUE #185: Editorial: There is no definition for "unclosed session"
This was originally noted in https://github.com/w3c/encrypted-media/pull/184#discussion-diff-62884149.

We use "unclosed session" three times in (normative parts of) the spec, but this term is not defined. In most `MediaKeySession` algorithms, we use language similar to  `the <a def-id="session-closed-algorithm"></a> algorithm has been run on this object`.

We should do one of the following:

1. Replace the "unclosed session" instances with such text.
1. Add a definition for "closed session", possibly using such text, and update all such text as well as these three cases to reference the definition.

(2) might be better from a readability perspective and avoids having to ensure we get the somewhat complex text correct everywhere.

mwatson2:


encrypted-media-respec.html
```diff
@@ -1606,22 +1606,26 @@
...
       <p>The MediaKeySession object represents a <a href="#key-session">key session</a>.</p>
       <p>
-        The User Agent SHALL execute the <a def-id="monitor-cdm-algorithm"></a> algorithm continuously for each <a>MediaKeySession</a> object until the sooner of
-        the <a def-id="session-closed-algorithm"></a> algorithm being run or the destruction of the <a>MediaKeySession</a>.
+        A <a>MediaKeySession</a> object is <dfn id="media-key-session-closed">closed</dfn> if and only if the <a def-id="session-closed-algorithm"></a> algorithm has
+        been run.
+      </p>
+      <p>
+        The User Agent SHALL execute the <a def-id="monitor-cdm-algorithm"></a> algorithm continuously for each <a>MediaKeySession</a> object
+        that is accessible to the page and that is not <a def-id="media-key-session-closed"></a>.
```
ddorwin: I'm not sure about the "accessible to the page" language. I'm not sure "destruction" is correct either (as it relates to garbage-collected objects), but I think we should stick with that until we get clarification otherwise.

encrypted-media-respec.html
```diff
@@ -1863,7 +1867,7 @@
                   </p>
                 </li>
                 <li><p>If the preceding step failed, or if <var>sanitized session ID</var> is empty, reject <var>promise</var> with a newly created <a def-id="TypeError"></a>.</p></li>
-                <li><p>If there is an unclosed session in this object's <a def-id="document-concept"></a> whose <a def-id="sessionId"></a> attribute is <var>sanitized session ID</var>, reject <var>promise</var> with a <a def-id="QuotaExceededError"></a>.</p>
+                <li><p>If there is a <a>MediaKeySession</a> object in this object's <a def-id="document-concept"></a> whose <a def-id="sessionId"></a> attribute is <var>sanitized session ID</var>, and that is not <a def-id="media-key-session-closed"></a>, reject <var>promise</var> with a <a def-id="QuotaExceededError"></a>.</p>
```
ddorwin: nit: I don't think the comma is correct. "and" probably isn't necessary either, but I understand the desire to break up the long statement.

ddorwin: It might read better to say "that is not closed" first. I didn't suggest this initially since the same session ID is the most relevant part, but the order is swapped this way at line 1883. I suggest using the same order in both places.

encrypted-media-respec.html
```diff
@@ -2690,9 +2694,10 @@
         <p><em>All</em> data associated with a session MUST be cleared when the session is cleared, such as in <a def-id="update"></a> when processing a license destruction acknowledgement or key usage record acknowledgement.
           See <a href="#persistent-state-requirements">Persistent Data</a>.
         </p>
-        <p>The CDM MUST ensure that data for a given session is only present in one active unclosed session in any <a def-id="document-concept"></a>.
+        <p>The CDM MUST ensure that data for a given session is only present in one <a>MediaKeySession</a> object in any <a def-id="document-concept"></a> that is not
```
ddorwin: "that" near the end of this line refers to the Document, which is not the intent.

encrypted-media-respec.html
mwatson2: If we write "each MediaKeySession object that has not been destroyed and ..." it reads a little strangely (what could it possible mean to run an algorithm on a destroyed object?).

So, perhaps the qualification is unnecessary and we can just say "each MediaKeySession object that is not closed" ?

PR #256, opened 2016-06-24
Fix #253: Specify Clear Key license release format

### LINKED ISSUE #253: Specify ClearKey License Release format for persistent-license sessions
Clear Key supports `temporary` and `persistent-usage-record` sessions, but not `persistent-license` sessions.

Since this is an optional capability, it does not have interop implications if added to the specification now.

The only additional material, aside from stating that this session type is supported, is the format for the record of license destruction, which could simply be an object containing the list of key ids.

mwatson2:


encrypted-media-respec.html
```diff
@@ -3532,17 +3546,34 @@
             Applications MAY decode the contents of the ArrayBuffer to a JSON string using the <a def-id="interface-textdecoder"></a> [[!ENCODING]].
           </p>
 
-          <section id="clear-key-release-format-example" class="informative">
-            <h5>Example</h5>
-            <p>The following example is a license release for a session which contained two keys. (Line breaks are for readability only.)</p>
+          <section id="clear-key-release-format-example-1" class="informative">
+            <h5>Example message reflecting a <var>record of key usage</var></h5>
```
ddorwin: I would reverse the examples to be consistent with the order above (simpler first).

encrypted-media-respec.html
```diff
@@ -3532,17 +3546,34 @@
...
+            <p>
+              The following example is a license release for a <a def-id="persistent-usage-record-session"></a> session which contained two keys.
```
ddorwin: s/which/that/. Same below.

ddorwin:
LGTM with comments. Thanks.

PR #258, opened 2016-06-24
Fix #243: Implementations should expose available modes that do not u…

### LINKED ISSUE #243: Privacy: Implementations that support modes with and without distinctive identifiers SHOULD expose this to the user
The spec already says:
>Implementations that use Distinctive Identifier(s) or Distinctive Permanent Identifier(s) SHOULD support the option to not use them.

I propose adding:
>Implementations with such support SHOULD expose the ability for the user to select this option.

ddorwin:
…se Distinctive Identifiers


mwatson2:
You seem to have picked up a bunch of ReSpec / stylesheet changes here too.

mwatson2:
Otherwise looks good to me.

PR #263, opened 2016-07-08
Fix #224: Define and user User Agent Profiles

### LINKED ISSUE #224: Use consistent terminology for browsing/user agent profile
Currently, the spec uses "profile," "browsing profile," and "user agent profile" to refer to the same thing. We should be consistent. I wonder if there is a standard term we should use and/or reference (like COOKIES).

mwatson2:


encrypted-media-respec.html
```diff
@@ -537,7 +537,18 @@
           </p>    
           <p>Time is intended to be consistent with Unix time.   
 		  </p>
-        </dd> 
+        </dd>
+        <dt id="user-agent-profile">User Agent Profile</dt>
```
ddorwin: Since this includes profiles in different OS user accounts, "browsing profile" might make more sense since it is not necessarily a feature of the UA.

encrypted-media-respec.html
```diff
@@ -537,7 +537,18 @@
...
+        <dd>
+          <p>
+            A User Agent on a given machine may support execution in a variety of different contexts or modes that are expected to behave independently in respect of page visible data.
```
ddorwin: Include temporary state (and nits): "with respect to application-visible state and data."

encrypted-media-respec.html
```diff
@@ -537,7 +537,18 @@
...
+            In particular, all stored data is expected to be independent. In this specification we refer to such independent contexts or modes as "User Agent Profiles".
+          </p>
+          <p>
```
ddorwin: Examples should be in a NOTE?

encrypted-media-respec.html
```diff
@@ -537,7 +537,18 @@
...
+            Examples of such independent contexts include if the user agent is running in different user accounts or if the user agent provides the capability to define
```
ddorwin: *operating system* user accounts...

encrypted-media-respec.html
```diff
@@ -4038,14 +4049,14 @@
...
           </p>
           <p>If information stored by one origin affects the operation of the Key System for another origin, then potentially the sites visited or content viewed by a user on one site may be revealed to another, potentially malicious, site.</p>
-          <p>If information stored for one browsing profile, browser, or user account on the client device affects the operation of the Key System for other profiles, browsers, or accounts, then potentially the sites visited or content viewed by in one may be revealed by or correlatable with another profile, browser, or account.</p>
+          <p>If information stored for one <a def-id="user-agent-profile"></a> on the client device affects the operation of the Key System for other <a def-id="user-agent-profile">user agent profiles</a>, or browsers, then potentially the sites visited or content viewed in one may be revealed by or correlatable with another <a def-id="user-agent-profile">user agent profiles</a> or browser.</p>
```
ddorwin: Although specifically mentioned in the definition of UAP,  I think the [operating system] user account part is especially relevant here and worth explicitly mentioning.

"Browsing profile" also seems particularly clearer here.

mwatson2:
"browsing profile" seems better than "user agent profile", but does it still carry an implication that this is a browser thing, not including user accounts ? Would "user profile" be better ?

(I agree with all the other comments)

encrypted-media-respec.html
```diff
@@ -3840,7 +3851,7 @@
             <li><p>Origins visited (via stored or in-memory data, permissions, etc.)</p></li>
             <li><p>Content viewed (via stored or in-memory licenses, keys, key IDs, records of key usage, records of license destruction, etc.)</p></li>
           </ul>
-          <p>This specification presents a specific concern because such information is commonly stored outside the user agent (and its profile storage), often in the CDM.</p>
+          <p>This specification presents a specific concern because such information is commonly stored outside the user agent (and its <a def-id="user-agent-profile"></a> storage), often in the CDM.</p>
```
ddorwin: Note: "its" here implies that whatever we define is a UA thing. Perhaps this should be "the," or "the associated," "the corresponding."

encrypted-media-respec.html
```diff
@@ -3786,7 +3797,7 @@
           <p>CDMs, especially those implemented outside the user agent, may not have the same fundamental isolations as the web platform.
             It is important that steps be taken to avoid information leakage, especially across origins.
             This includes both in-memory and stored data.
-            Failure to do so could lead to information leakage to/from private browsing sessions, across profiles, and even across different browsers, applications, and operating system user accounts.
+            Failure to do so could lead to information leakage to/from private browsing sessions, across <a def-id="user-agent-profile"></a>, and even across different browsers, applications, and operating system user accounts.
```
ddorwin: I think this one should be plural.

ddorwin:
I think we do want to imply that it is related to the "profile" used by the browser/user agent. I meant that this profile could be the product of multiple things (e.g. UA profile feature + OS user account) so saying it is just the "user agent profile" seemed inaccurate. "User profile" seems too vague and could be interpreted as just the OS account.

encrypted-media-respec.html
```diff
@@ -4038,14 +4039,14 @@
...
           </p>
           <p>If information stored by one origin affects the operation of the Key System for another origin, then potentially the sites visited or content viewed by a user on one site may be revealed to another, potentially malicious, site.</p>
-          <p>If information stored for one <a def-id="user-agent-profile"></a> on the client device affects the operation of the Key System for other <a def-id="user-agent-profile">user agent profiles</a>, or browsers, then potentially the sites visited or content viewed in one may be revealed by or correlatable with another <a def-id="user-agent-profile">user agent profiles</a> or browser.</p>
+          <p>If information stored for one <a def-id="browsing-profile"></a> on the client device affects the operation of the Key System for other <a def-id="browsing-profile">browsing profiles</a>, or browsers, then potentially the sites visited or content viewed in one may be revealed by or correlatable with another <a def-id="browsing-profile">browsing profiles</a> or browser. Note that different operating system user accounts are defined as different <a def-id="browsing-profile">browsing profiles</a>.</p>
```
ddorwin: The BP instance after "another" should be singular.

ddorwin: While accurate, the "Note..." sentence seems longer and less precise than necessary. Does adding ", including across operating system user accounts" to the previous sentence work?

ddorwin:
Thanks. LGTM with this singular comment addressed. I have another suggestion too, but I'll leave that to you.

encrypted-media-respec.html
mwatson2: I was concerned that the presence of the "including ..." qualifier here, but not in other places, might be mis-interpreted as meaning that OS accounts were included only in this case and not in others.

encrypted-media-respec.html
```diff
@@ -4039,7 +4039,7 @@
...
-          <p>If information stored for one <a def-id="browsing-profile"></a> on the client device affects the operation of the Key System for other <a def-id="browsing-profile">browsing profiles</a>, or browsers, then potentially the sites visited or content viewed in one may be revealed by or correlatable with another <a def-id="browsing-profile">browsing profiles</a> or browser. Note that different operating system user accounts are defined as different <a def-id="browsing-profile">browsing profiles</a>.</p>
+          <p>If information stored for one <a def-id="browsing-profile"></a> on the client device affects the operation of the Key System for other <a def-id="browsing-profile">browsing profiles</a>, or browsers, then potentially the sites visited or content viewed in one may be revealed by or correlatable with another <a def-id="browsing-profile">browsing profile</a> or browser. Note that different operating system user accounts are defined as different <a def-id="browsing-profile">browsing profiles</a>.</p>
```
ddorwin: nit: You don't need the "browsing profile" text anymore.

encrypted-media-respec.html
ddorwin: I'm not too concerned about that, and line 3790, which is also trying to warn about possible issues, has "including" language.

This case also has "or browser" so it's not just the profile. Does "even across..." work better?

My main concern, though, is that the "Note..." text does not clearly state the implications or even make it clear why this is relevant to this paragraph. I'm more concerned about that than misinterpreting the term, which is clearly defined.

ddorwin:
LGTM

PR #265, opened 2016-07-08
Fix #262: Clarify meaning of the combination of requested media capab…

### LINKED ISSUE #262: Steps regarding satisfying robustness requirements in "Get Supported Capabilities for Audio/Video Type" algorithm can be misinterpreted
The "Get Supported Capabilities for Audio/Video Type" algorithm as specified can be interpreted in several ways. The language needs to be tightened up.

In step 13 of the algorithm, it states:

>13. If the user agent and implementation definitely support playback of encrypted media data for the combination of container, media types, robustness and local accumulated configuration in combination with restrictions:
>NOTE  requested media capability (content type and robustness) must be supported with all previously added requested media capabilities.
>> 1. Add requested media capability to supported media capabilities. 

...and in step 2 after adding the requested media capability to the local accumulated configuration, the algorithm states:

>> NOTE  This step ensures that configurations are always checked with configurations from previous iterations, including from previous calls to this algorithm. Otherwise, only configurations from previous calls to this algorithm would be checked in subsequent calls.

It seems to me that one way to read that would be that all robustness requirements in previously added MediaKeySystemMediaCapabilitys must be also be satisfied in any new MediaKeySystemMediaCapabilitys that we add. That means we'd need to apply video restrictions to audio for example, which seems wrong.

For example, if we've added a video capability with secure-decode robustness, then we considered adding an audio capability with only secure-decrypt capability, we'd be obliged to refuse to add the audio capability, since secure-decrypt is not as robust as secure-decode.

Another way to read this would be that every MediaKeySystemMediaCapability that we add must satisfy its own robustness requirements, and that implies that all added MediaKeySystemMediaCapabilitys satisfy their requirements. But if that was the interpretation, then why would we need the _local accumulated configuration_?

Or another interpretation could be that all previously added MediaKeySystemMediaCapability's robustness requirements can be supported _when used together with_ the MediaKeySystemMediaCapability being considered for addition?

That last interpretation I think would make the most sense. The wording should be made clearer here.

mwatson2:
…ilities with previously added instances of the same

### DIFF
--- a/encrypted-media-respec.html
+++ b/encrypted-media-respec.html
@@ -1013,7 +1013,7 @@
                   </li>
                   <li><p>If <var>robustness</var> is not the empty string and contains an unrecognized value or a value not supported by <var>implementation</var>, continue to the next iteration. String comparison is case-sensitive.</p></li>
                   <li><p>If the user agent and <var>implementation</var> definitely support playback of encrypted <a def-id="media-data"></a> for the combination of <var>container</var>, <var>media types</var>, <var>robustness</var> and <var>local accumulated configuration</var> in combination with <var>restrictions</var>:</p>
-                    <p class="note"><var>requested media capability</var> (content type and robustness) must be supported with all previously added requested media capabilities.</p>
+                    <p class="note"><var>requested media capability</var> (content type and robustness) must be supported when used together with all previously added requested media capabilities.</p>
                     <ol>
                       <li>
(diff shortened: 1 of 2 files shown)


ddorwin:
The new text LGTM. I'll defer to @cpearce as to whether this addresses the issue, but this PR is at least an improvement.

PR #267, opened 2016-07-12
Fix #227: Simplify setting of _use distinctive identifier_ and _persi…

### LINKED ISSUE #227: Simplify logic for setting <var>use distinctive identifier</var> and <var>persistent state allowed</var>
Rather than explicitly checking for `"not-allowed"` and having a note that the value cannot be `"optional"` in this algorithm, we could just check for "is `"required"` and "Otherwise." This is also more robust against breaking than a non-normative note.

mwatson2:
…stent state allowed_


ddorwin:
LGTM

Issue #269, opened 2016-07-14
Secure origin check should not use "incumbant" settings object


mwatson2:
Things have changed: "current settings object" is apparently now the right thing:

https://readable-email.org/list/public-script-coord/topic/multiple-globals-and-you

https://github.com/whatwg/html/issues/1430

@dontcallmedom

ddorwin:
From #270:
>https://w3c.github.io/webappsec-secure-contexts/#new has updated the recommended language. See https://readable-email.org/list/public-script-coord/topic/multiple-globals-and-you for details.

In addition, there is an alternative/additional recommendation to use `[SecureContext]` in the IDL. (There is a question about whether to use one or both methods in https://github.com/w3c/webcrypto/issues/28#issuecomment-232805001.)

If we maintain the algorithm check, we may want to reject the promise rather than `NotSupportedError` with `SecurityError` as suggested in the recommended language.

ddorwin:
Per https://github.com/w3c/webcrypto/issues/28#issuecomment-232857255, `[SecureContext]` in the IDL is sufficient.

ddorwin:
#271 uses `[SecureContext]`, which allowed us to eliminate the text covered by this issue, making it obsolete.

PR #271, opened 2016-07-15
Fix #220: Remove deprecated support for insecure contexts

### LINKED ISSUE #220: Resolve question of support for insecure contexts
At the April 2015 face-to-face, we decided to mark support for insecure contexts as DEPRECATED and reevaluate support during CR. There is a Note to this effect in the [`requestMediaKeySystemAccess()` algorithm](https://w3c.github.io/encrypted-media/#requestMediaKeySystemAccess). We either need to remove the `Allowed` step and Note or otherwise update the Note before PR.

ddorwin:
Also updated text related to use of the APIs on insecure contexts.

The algorithm step is replaced with the `[SecureContext]` IDL attribute as discussed in #269.


joeyparrish:
On what timeline will Chrome start enforcing secure origin?  When the info is available, I'd like to announce to the shaka-player-users list the date or Chrome release where this will happen.

ddorwin:
@joeyparrish: Support on insecure contexts has been deprecated in Chrome for over a year, and we have seen content providers switching to HTTPS. While we have not yet given a timeline for enforcement, I expect that to happen soon, and we will clearly communicate this in Chrome and via other channels. We believe this spec change is an important step in presenting a consistent message - and eventually behavior - to authors as well as implementers.

steelejoe:
Changes look good.

mwatson2:
Looks good to me.

PR #277, opened 2016-07-19
Fix #275: generateRequest should use var message type for the message…

### LINKED ISSUE #275: generateRequest may generate incorrect message type
The steps of `generateRequest` set _message_ and _message type_ to the message to be generated and its type. However, step 10.9.4 always uses `license-request` as the message type, instead of _message type_.

mwatson2:
… type


ddorwin:
LGTM

PR #282, opened 2016-07-20
Fix #279: Fix method links after Contiguous IDL migration

### LINKED ISSUE #279: Method name anchors are broken
I had added code to `encrypted-media.js` to generate simple stable anchors for methods that used exactly the method name. For example, https://w3c.github.io/encrypted-media/#generateRequest. These are used in bug reports, code comments, etc.

After #274, these no longer work. The code in `encrypted-media.js` probably needs to be updated.

mwatson2:


ddorwin:
LGTM. Thank you.

PR #285, opened 2016-08-01
Fix #153: Remove "run when" text from HTMLMediaElement algorithms

### LINKED ISSUE #153: Non-method algorithm intro paragraphs may be misleading about when to run the algorithm

ddorwin:
Also fixed up text related to queueing of "encrypted" and "message" events.

### DIFF
--- a/encrypted-media-respec.html
+++ b/encrypted-media-respec.html
@@ -2352,7 +2352,7 @@ interface MediaKeyMessageEvent : Event {
           <ol>
             <li><p>Let the <var>session</var> be the specified <a>MediaKeySession</a> object.</p></li>
             <li>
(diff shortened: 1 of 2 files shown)


mwatson2:
Looks good to me.

PR #287, opened 2016-08-02
Fix #223: Identifiers MUST NOT be shared across browsing profiles

### LINKED ISSUE #223: Privacy: Clarify "distinctive identifier" around multiple users.
In the spec about "Non-Associable Identifiers", it says:

> Implementation must not expose associable identifiers to applications, even in encrypted form, across origins or across clearing of identifiers.

Other than across "origins" and across "clearing of identifiers", there is another case which could be sensitive: across "multiple users". For example, multiple OS user accounts, or multiple user profiles (if supported by the user agent) . It feels bad if a JS application (origin) can associate different users using the same device through the distinctive identifier.

Should we add some requirement or recommendation that "implementation should not expose associable identifiers to applications across different users"?

ddorwin:


mwatson2:
Looks good to me.

PR #290, opened 2016-08-03
Consent MUST be per-browsing profile

ddorwin:
Also cleans up text related to consent and network attacks.

### DIFF
--- a/encrypted-media-respec.html
+++ b/encrypted-media-respec.html
@@ -3611,6 +3611,12 @@ interface MediaEncryptedEvent : Event {
                 <li><p>Data, including <a def-id="distinctive-identifiers"></a> and <a def-id="distinctive-permanent-identifiers"></a>, transmitted between the client and server can be viewed and/or modified by other entities. See <a href="#user-tracking">User Tracking</a>.</p></li>
               </ul>
             </li>
+            <li>
+              <p>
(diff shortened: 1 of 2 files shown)


ddorwin:
This addresses some things I found while working on other PRs.

ddorwin:
This is also consistent with issue #223 and PR #287.

encrypted-media-respec.html
```diff
@@ -3611,6 +3611,12 @@
                 <li><p>Data, including <a def-id="distinctive-identifiers"></a> and <a def-id="distinctive-permanent-identifiers"></a>, transmitted between the client and server can be viewed and/or modified by other entities. See <a href="#user-tracking">User Tracking</a>.</p></li>
               </ul>
             </li>
+            <li>
```
jdsmith3000: I don't fully understand the intent of this addition.  It doesn't seem related to browsing profiles, but perhaps does provide an additional threat that warrants secured origins.  Is that the purpose?

ddorwin: I moved this from line 3877, which I replaced with a more general reference to this section.

jdsmith3000: I see the original comment, and agree this section is more appropriate.  LGTM.

PR #291, opened 2016-08-03
Fix #283: Replace "these APIs" with "the APIs defined in this specification"

### LINKED ISSUE #283: “these APIs” is ambiguous
The spec refers to the APIs defined in the EME specification as these APIs 22 times. While longer, we may want to say “the APIs [defined] in this specification” instead.

As an example, in the following statement from https://w3c.github.io/encrypted-media/#direct-individualization, a section outside any of the objects, it might not be clear which APIs we are referring to. (We’re mostly referring to “message” events and similar.)
>MUST NOT be passed to or through the application via these APIs.

ddorwin:


ddorwin:
This builds on #290, which is the first commit. Until #290 is landed and merged into this branch, just review the second commit, which is the real difference for this PR.

jdsmith3000:
The API clarifications in the second commit LGTM.

PR #294, opened 2016-08-03
Fix #289: Restrict nearly all interfaces to secure contexts

### LINKED ISSUE #289: Restrict all interfaces to secure contexts
#271 added `[SecureContext]` to `requestMediaKeySystemAccess()`, which effectively makes use of the APIs and CDM impossible on insecure contexts. However, all the other dictionaries, interfaces, and partial interfaces will still be visible to insecure contexts. This is a bit strange since feature detection using `if (MediaKeys)` would succeed while `if (navigator.requestMediaKeySystemAccess)` would not.

Probably the correct/consistent thing to do would be to add `[SecureContext]` to all IDL blocks, including `partial interface HTMLMediaElement`. This is easy and straightforward.

However, that alone would not prevent the `encrypted` and `waitingforkey` events from being fired even though the `on...` attributes would not be usable. I see three options:
 1. Abort the relevant algorithms when in an insecure context, preventing the events from being fired.
 1. Fire the events but leave the `on...` attributes inaccessible.
 1. Fire the events and allow use of the `on...` attributes but leave everything else inaccessible. (This is accomplished by adding `[SecureContext]` to the `mediaKeys` and `setMediaKeys` members rather than the entire `partial interface`.)

(2) and (3) allow players on insecure contexts to detect but not handle encrypted content, which seems good. (2) seems like unexpected behavior and limits the benefits in the previous sentence. Therefore, I am leaning towards (3).

@mikewest, any comments on the preferred behavior?

ddorwin:
Only the "encrypted" and "waitingforkey" events are exposed on insecure contexts.

### DIFF
--- a/encrypted-media-respec.html
+++ b/encrypted-media-respec.html
@@ -611,7 +611,7 @@
               Otherwise, it is resolved with a new <a>MediaKeySystemAccess</a> object.
             </p>
             <div class="note">
-              <p>This method is only exposed to <a href="https://www.w3.org/TR/secure-contexts/#secure-context">secure contexts</a> [[POWERFUL-FEATURES]] as indicated by the <code>[SecureContext]</code> IDL attribute.</p>
+              <p>This method is only exposed to <a href="https://www.w3.org/TR/secure-contexts/#secure-context">secure contexts</a> [[SECURE-CONTEXTS]] as indicated by the <code>[SecureContext]</code> IDL attribute.</p>
               <p>
                 Requiring Secure Contexts is <em>not</em> a replacement for other security- and privacy-related requirements and recommendations.
                 Implementations MUST meet all related requirements and SHOULD follow related recommendations such that the risks on in an secure context would be similar.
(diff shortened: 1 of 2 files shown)


encrypted-media-respec.html
```diff
@@ -611,7 +611,7 @@
               Otherwise, it is resolved with a new <a>MediaKeySystemAccess</a> object.
             </p>
             <div class="note">
-              <p>This method is only exposed to <a href="https://w3c.github.io/webappsec-secure-contexts/#secure-context">secure contexts</a> [[POWERFUL-FEATURES]] as indicated by the <code>[SecureContext]</code> IDL attribute.</p>
+              <p>This method is only exposed to <a href="https://www.w3.org/TR/secure-contexts/#secure-context">secure contexts</a> [[POWERFUL-FEATURES]] as indicated by the <code>[SecureContext]</code> IDL attribute.</p>
```
mwatson2: I believe the database is now updated so you can write `[[SECURE-CONTEXTS]]`.

mwatson2:
LGTM

Suggest s/[[POWERFUL-FEATURES]]/[[SECURE-CONTEXTS]]/

encrypted-media-respec.html
ddorwin: Done.

PR #295, opened 2016-08-04
Fix #248: Provide references to distinctiveIdentifier in relevant sections

### LINKED ISSUE #248: Explicitly state that distinctive and permanent identifiers may not be used without distinctiveIdentifier being "required"
We have algorithmic steps that say whether distinctive and permanent identifiers may be used, but I don't believe we have a clear statement in the definition of these identifiers that says they are only accessible when the MediaKeySystemAccess has `distinctiveIdentifier="required"`.

Even a non-normative note clearly stating this might be helpful for the reader.

It might also be helpful to include this in places that reference these types of identifiers, such as the individualization section.

ddorwin:


ddorwin:
Note that the new text in the DI section uses capital case, which is consistent with #296.

encrypted-media-respec.html
```diff
@@ -441,6 +445,10 @@
               (The <a def-id="distinctive-permanent-identifier-maybe-plural"></a> MUST NOT ever be provided to the application, even in opaque or encrypted form.) 
             </p>
           </div>
+          <p class="note">
+            <a def-id="option-distinctiveIdentifier"></a> controls whether Distinctive Identifiers may be used.
+            Specifically, Distinctive Identifiers may only be used when the value of the <a def-id="option-distinctiveIdentifier"></a> member of the <a>MediaKeySystemAccess</a> used to create the <a>MediaKeys</a> object is <a def-id="requirement-required"></a>.
+          </p>
```
mwatson2: I don't really agree that it's a good idea to duplicate this, although I appreciate it's a note and not a requirement. I think we should find the most appropriate place and include the note there, once.

jdsmith3000: My first instinct was to agree with this comment, but the notes explicitly apply the requirement for Distinctive Identifiers, Distinctive Permanent Identifiers, and user of identifiers.  To me, the current approach seems preferred.

ddorwin:
@jdsmith3000, to clarify, does "current approach" mean the current state of this PR (meaning you agree with the change)?

@mwatson2, as @jdsmith3000 notes, the notes are added to the definitions that are affected by the option. I don't think there is one place that is more appropriate than the others, which is probably an argument for including it in each.

jdsmith3000:
Correct.  The change as is (4 similar notes) LGTM.

PR #296, opened 2016-08-04
Capitalize instances of Distinctive Identifier in the section that defines it

ddorwin:
This is consistent with other uses and definitions. I don't remember why these weren't capitalized before.


jdsmith3000:
LGTM.  I tried looking for any missed instances and couldn't find any.

Issue #293, opened 2016-08-03
WebIDL has link errors after contiguous IDL conversion


jdsmith3000:
EME WebIDLs has some link errors:

- `Interface MediaKeys` has a link, but shouldn’t.  It’s linked to the `mediaKeys` attribute for HTMLMediaElements.

- The HTMLMediaElement `mediaKeys` attribute doesn’t have a link, and `MediaKeys?` is incorrectly linked to the attribute instead of the enum.

@mwatson2 reports that this is the same ReSpec bug that caused trouble for MSE (w3c/media-source#83): two things with the same (case-insensitive) name.

EME also doesn't have the links to external data types that MSE has (looks like it never had them).

He can do the same fix as I did for MSE, or we can wait and see if the ReSpec bug gets fixed in time.

ddorwin:
* All MediaKeys links (`<a>MediaKeys</a>` in the source) reference the member.
  * Including in the attribute definition:  `mediaKeys of type MediaKeys`

While w3c/media-source#83 doesn't reference this issue, the PR for it (w3c/media-source#127) does. Most relevant, though, is the ReSpec bug: https://github.com/w3c/respec/issues/893.

ddorwin:
PR #299 works around this until w3c/respec#893 is fixed.

ddorwin:
The issues identified above should be fixed (please verify), but the solution is not ideal. I think we've done all we can until w3c/respec#893 is fixed and we can remove the workaround (see the references to `https://github.com/w3c/respec/issues/893` in the `-respec.html` source).

ddorwin:
To be addressed in the future.

PR #297, opened 2016-08-08
Issue #293: Work around w3c/respec#893

ddorwin:
Some links are removed, but there should not be any incorrect ones.


ddorwin:
Replaced by #299, which is identical

PR #298, opened 2016-08-08
Fix references to MediaEncryptedEvent members

ddorwin:
They macros previously referenced the MediaEncryptedEventInit members.


ddorwin:
This is editorial, so I'm going to commit.

Issue #301, opened 2016-08-09
Move the Individualization section out of the Privacy section


ddorwin:
Individualization is currently a subset of the Privacy User Tracking section, but it is more than just a privacy topic. There is a TODO in the source related to this.

There isn't an obvious place for it, but Implementation Requirements (after the Identifiers subsection) is probably better. We can reference the requirements in that section from the User Tracking section.

steelejoe:
I agree, the Implementation Requirements section is a better fit. This could follow the model of "Support Multiple Keys" e.g. "Implementations must support Key Systems which send the _individualization-request_ message type."

mwatson2:
Makes sense.

ddorwin:
PR #306.

PR #303, opened 2016-08-12
Fix #273: Address comments on security and privacy sections

### LINKED ISSUE #273: Comments on security and privacy sections
@steelejoe wrote the following in https://github.com/w3c/encrypted-media/issues/221#issuecomment-224341040. I've moved the remaining items (and one reply) here to separate them from the larger "review" issue #221.

>I have some nit-picky comments. Should be relatively easy to address. 
>
>Section 10.3.2 Mitigations
TLS -- "Furthermore, origin-specific permissions in combination with a secure origin, **ensure** that permissions granted to an application **cannot be abused** by a network attacker.". This language is too strong. We can safely say that TLS makes abuse far less likely, but not impossible. This should be changed to "Origin-specific permissions in combination with a secure origin make abuse of permissions granted to an application by a network attacker far less likely.".
>...
>
>Section 11.4.2 Mitigations "Shared blacklists"
This section is unclear. What is a "Key System origin"? Does this refer to the dotted identifier for the Key System itself, in which case a blacklist seems like overkill since there are only a small number of Key Systems. Or does this refer specifically to use a particular Key System on an application origin? The latter seems more likely. If that is the correct interpretation this should be changed to "User agents may allow users to share blacklists of application origins and/or Key Systems". 
>
>Section 11.4.2 Mitigations "Per-origin user alerts / prompts and permissions"
This sentence "User agents must prompt or otherwise inform the user before allowing use of a Distinctive Identifier that is not unique per-origin and/or not clearable is used. " is unnecessary based on the note following. Why not just remove it and the note? Apologies if this has been discussed in depth -- I did not see it.

ddorwin:


steelejoe:
This change looks good. I am still bothered by the TLS text. However If I can't propose something better in the next day or so I will let it go.

ddorwin:
@steelejoe, this is the last PR blocking #301 (due to conflicting changes). Can we land this then reopen #273 if you come up with a proposal?

steelejoe:
Please go ahead and land the changes.

PR #306, opened 2016-08-16
Fix #301: Move Individualization to Requirements section

### LINKED ISSUE #301: Move the Individualization section out of the Privacy section

ddorwin:
Moves the section and updates the anchor name. No content changes.


ddorwin:
If there are no comments, I'll land this soon so we can move forward on other items.

mwatson2:
Looks good to me.

PR #307, opened 2016-08-30
Fix #237: Update User Tracking section to address Distinctive Permanent Identifiers

### LINKED ISSUE #237: Privacy: Update User Tracking section, including to specifically address Permanent Identifiers
A lot has changed with respect to Distinctive Identifiers and related issues since the [User Tracking](https://w3c.github.io/encrypted-media/#user-tracking) section was written. We should review and update it to be consistent. Specifically, [Distinctive] Permanent Identifiers should be explicitly addressed as these pose a different set of tracking concerns.

ddorwin:
Also reference the concerns and mitigations related to individualization, which
was extracted from this section for issue #301.

### DIFF
--- a/encrypted-media-respec.html
+++ b/encrypted-media-respec.html
@@ -3876,7 +3876,7 @@ interface MediaEncryptedEvent : Event {
           </p>
           
           <p>
-            In addition to the potential for sites and other third-parties to track users, the user agent and/or CDM implementer could build a profile of the user's activities or interests, such as sites using the APIs defined in this specification that the user visits.
+            In addition to the potential for sites and other third-parties to track users, the user agent implementer, CDM vendor, or device vendor could build a profile of the user's activities or interests, such as sites using the APIs defined in this specification that the user visits.
             Such tracking would undermine the privacy protections provided by the rest of the web platform, especially those related to isolation of origins.
           </p>
           
--- a/encrypted-media-respec.html
+++ b/encrypted-media-respec.html
@@ -3881,7 +3881,7 @@ interface MediaEncryptedEvent : Event {
           </p>
           
           <p>
-            Identifiers, such as <a def-id="distinctive-identifiers"></a>, are often obtained from a server operated or provided by the CDM vendor, such as via an <a href="#individualization">individualization</a> process.
+            Identifiers, such as <a def-id="distinctive-identifiers"></a>, may be obtained from a server operated or provided by the CDM vendor, such as via an <a href="#individualization">individualization</a> process.
             The process may include providing client identifier(s), including <a def-id="distinctive-permanent-identifier-maybe-plural"></a>, to the server.
             In order to generate a per-origin identifier, a value representing the origin may also be provided.
           </p>
@@ -3906,9 +3906,9 @@ interface MediaEncryptedEvent : Event {
             <dt>Use origin-specific and browsing profile-specific Key System storage</dt>
             <dd>
               <!-- TODO: Reference section added for issue #242. -->
-              <p>Any data persistently stored by the CDM that might impact messages or behavior in an application- or license server-visible way MUST be stored in an <a def-id="origin"></a>-specific and <a def-id="browsing-profile"></a>-specific way and MUST NOT leak to or from private browsing sessions.
-                Session data, licenses, keys, and other data that are persistently stored MUST be stored per-<a def-id="origin"></a> and per-<a def-id="browsing-profile"></a>.
-                See <a def-id="session-storage"></a>.
+              <p>Any data used by the CDM that might impact messages or behavior in an application- or license server-visible way MUST be partitioned by <a def-id="origin"></a> and <a def-id="browsing-profile"></a> and MUST NOT leak to or from private browsing sessions.
+                This includes both in-memory and persisted data.
+                Specifically but not exhaustively, session data, licenses, keys, and per-origin identifiers MUST be partitioned per-<a def-id="origin"></a> and per-<a def-id="browsing-profile"></a>.
               </p>
             </dd>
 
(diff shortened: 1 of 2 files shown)


encrypted-media-respec.html
```diff
@@ -3868,23 +3870,48 @@
...
+          <p>In addition, if a Key System permits keys or other data to be stored and to be re-used between origins, then it may be possible for two origins to collude and track a unique user by recording their ability to access a common key.</p>
           <p>Finally, if any user interface for user control of Key Systems presents data separately from data in HTTP session cookies [[COOKIES]] or persistent storage, then users are likely to modify site authorization or delete data in one and not the others.
             This would allow sites to use the various features as redundant backup for each other, defeating a user's attempts to protect his or her privacy.
           </p>
+          
+          <p>
+            In addition to the potential for sites and other third-parties to track users, the user agent and/or CDM implementer could build a profile of the user's activities or interests, such as sites using the APIs defined in this specification that the user visits.
+            Such tracking would undermine the privacy protections provided by the rest of the web platform, especially those related to isolation of origins.
```
mwatson2: The ability of the user agent to build a profile of the users' activities is not modified by this specification.

It's not clear how the CDM implementor could do this if they are compliant to the requirements to isolate data by origin.

ddorwin: >The ability of the user agent to build a profile of the users' activities is not modified by this specification.

Agreed. I was trying to be inclusive since the user agent vendor could be the one receiving individualization requests. For the user agent vendor, it the difference would be addition of _remote_ tracking. In contrast, local tracking by a user agent would be very concerning and more obvious.

I think this text is fine, especially given the context, but I'm open to suggestions.

mwatson2: I guess my point was that if you are worried about the User Agent implementor tracking users, you have much bigger problems than EME. The User Agent could simply radio back home with information on every site you visit and every UI element you click, every password you enter etc. This specification does not make that any worse.

I fact some User Agent implementors *do* build up a cloud-stored profile, but it is used only for the benefit of the user, to make preferences, bookmarks etc. available across devices and only done with user consent (I assume).

So, I'd suggest removing "user agent and/or".

encrypted-media-respec.html
```diff
@@ -3868,23 +3870,48 @@
...
           </p>
+          
+          <p>
+            In addition to the potential for sites and other third-parties to track users, the user agent implementer, CDM vendor, or device vendor could build a profile of the user's activities or interests, such as sites using the APIs defined in this specification that the user visits.
+            Such tracking would undermine the privacy protections provided by the rest of the web platform, especially those related to isolation of origins.
+          </p>
+          
+          <p>
+            Identifiers, such as <a def-id="distinctive-identifiers"></a>, are often obtained from a server operated or provided by the CDM vendor, such as via an <a href="#individualization">individualization</a> process.
+            The process may include providing client identifier(s), including <a def-id="distinctive-permanent-identifier-maybe-plural"></a>, to the server.
```
mwatson2: I don't think we know very much about how people are doing individualization, so rather than "are often obtained" perhaps we should just say "may be obtained" (unless you have direct knowledge of people often doing it this way).

encrypted-media-respec.html
```diff
@@ -3868,23 +3870,48 @@
...
                 See <a href="#limit-or-avoid-use-of-distinctive-identifiers-and-permanent-identifiers">Limit or Avoid use of Distinctive Identifiers and Permanent Identifiers</a>.
               </p>
             </dd>
 
+            <dt>Use origin-specific and browsing profile-specific Key System storage</dt>
+            <dd>
+              <!-- TODO: Reference section added for issue #242. -->
+              <p>Any data persistently stored by the CDM that might impact messages or behavior in an application- or license server-visible way MUST be stored in an <a def-id="origin"></a>-specific and <a def-id="browsing-profile"></a>-specific way and MUST NOT leak to or from private browsing sessions.
+                Session data, licenses, keys, and other data that are persistently stored MUST be stored per-<a def-id="origin"></a> and per-<a def-id="browsing-profile"></a>.
```
mwatson2: Is this new text or moved text ?

It seems a little too broad. CDMs may store some kind of origin-independent identity which is used to generate origin-specific, non-associable identifiers

encrypted-media-respec.html
```diff
@@ -3900,6 +3927,21 @@
...
+              </p>
+            </dd>
+
+            <dt>Do not expose <a def-id="distinctive-permanent-identifiers"></a> to the application</dt>
+            <dd>
+              <p>
+                Implementations MUST NOT expose <a def-id="distinctive-permanent-identifiers"></a> to the application or origin. 
+              </p>
+            </dd>
+
```
mwatson2: Is this new or moved ?

mwatson2:
This looks good aside from my last two comments. The first I think needs to be addressed. For the second, if it's existing text it's fine, but if it's new text, whilst I don't disagree with it, I think it should be discussed to ensure we have wide agreement (the first part, not the part about permanent identifiers, which is specified elsewhere already, I believe).

mwatson2:
In general, I don't think we do a good job of describing which of the things in the "Concerns" section are mitigated, and to what extent, by the things in the mitigations section. This is in part because we started with the concerns and designed the mitigations to address them.

So, a reader of the specification may be somewhat horrified reading the "Concerns" without realizing that they have been addressed.

Ideally, we would put the requirements first, follow this by any concerns which are not fully addressed (i.e. new concerns that this specification raises) and move remaining material into a "motivations" section. Perhaps this is a task for V2.

encrypted-media-respec.html
ddorwin: This entire section was copied from line 4031 below with one exception - it looks like I added "other data" since this was about tracking in general.

I believe the second sentence is supposed to reinforce the first one. I changed it to note that these are specific examples.

encrypted-media-respec.html
ddorwin: The first item is restating the requirements in the Individualization section.

The second item is restating what is said in the DPI definition:
>Distinctive Permanent Identifiers must not ever be exposed to the application, even in encrypted form.

ddorwin:
I agree we could do a better job indicating which concerns are mitigated for all compliant implementations. Still, I think it is worth identifying the concerns exposing such capabilities could present, but we should be clear that this is the list of possible concerns that drove the design.

For this PR in particular, many of the items are completely addressed. However, as a counterexample, there is at least some information allowed to leak in the individualization process even when all requirements and recommendations are followed.

encrypted-media-respec.html
mwatson2: Ok

encrypted-media-respec.html
mwatson2: Ok.

mwatson2:
LGTM

"keys" in the section I commented on could be linked to the definition, but that is cosmetic.

PR #309, opened 2016-08-30
Fix #308: Non-distinctive Permanent Identifiers may be exposed to the application or origin

### LINKED ISSUE #308: Non-distinctive Permanent Identifiers may be exposed to the application or origin
The NOTE in [Use Per-Origin Per-Profile Identifiers](https://w3c.github.io/encrypted-media/#per-origin-per-profile-identifiers) currently [1] says:
>Permanent Identifiers MUST NOT be exposed to the application or origin.

This statement includes all permanent identifiers, including values that are shared by, for example, all users of a specific device model [2] or platform. This is _not_ the intent and would likely prohibit any implementation that does not use Distinctive Identifiers.

Looking at the the commit that [added](https://github.com/w3c/encrypted-media/commit/765295425f6fa259c76eea1169d24413268fce51#diff-f72607e47a6f74e53dc90eab8ee094e2R3262) this text, it appears this statement was intended to address the exception for Permanent Identifiers being per-origin in the normative text. However, I believe this should have been **Distinctive** Permanent Identifiers.

The lack of requirements around non-distinctive Permanent Identifiers, that is Permanent Identifiers that are not Distinctive Permanent Identifiers, would seem to indicate there are not significant concerns about these, and, as mentioned above, they are essentially required for implementations that avoid using user-specific Distinctive Identifiers.

[1] This is just used as a possible example. The spec does not say whether such a model key is distinctive. The spec says, "A Distinctive Permanent Identifier is a Permanent Identifier that is _not_ shared across a large population of users or client devices," but "large population" is not defined. This bug and example use here does not change that.

ddorwin:
The previous text covering all Permanent Identifiers was too broad.

### DIFF
--- a/encrypted-media-respec.html
+++ b/encrypted-media-respec.html
@@ -3223,7 +3223,7 @@ interface MediaEncryptedEvent : Event {
           </p>
           <div class="note">
             <p>This includes but is not limited to <a def-id="distinctive-identifiers"></a>.</p>
-            <p><a href="#permanent-identifier">Permanent Identifiers</a> MUST NOT be exposed to the application or origin.</p>
+            <p><a def-id="distinctive-permanent-identifiers"></a> MUST NOT be exposed to the application or origin.</p>
           </div>
           <p>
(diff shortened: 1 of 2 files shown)


mwatson2:
LGTM

jdsmith3000:
LGTM also.

PR #318, opened 2016-09-08
Fix #313: Add step to allow user agent to return InformUser for reaso…

### LINKED ISSUE #313: Add step to allow user agent to return InformUser for reasons other than use of Distinctive Identifiers
The [Get Consent Status](https://w3c.github.io/encrypted-media/#get-consent-status) algorithm currently allows "The user agent [to require] explicit user consent for the _accumulated configuration_ for other reasons." (This was added for issue #96.) However, there is no such option for the user agent to return `InformUser` unless "the `distinctiveIdentifier` member of accumulated configuration is not "not-allowed"."

We should add a new step after the above text that says:
>If the user agent requires informing the user for the _accumulated configuration_ for other reasons, return `InformUser`.

This is consistent with the following statements from the Security section:
>If a user agent chooses to support a Key System implementation that cannot be sufficiently sandboxed or otherwise secured, the user agent SHOULD ensure that users are fully informed and/or give explicit consent before loading or invoking it.

<p/>
>User agents SHOULD ensure that users are fully informed and/or give explicit consent before a Key System that presents security concerns that are greater than other user agent features (e.g. DOM content) may be accessed by an origin.

Note: #312 could cause more conditions to be added to the existing step that requires returning `InformUser`. In that case (or even regardless), we could restructure the existing step to match the "If any of the following are true:" structure of the text added for #96.

ddorwin:
…ns other than use of Distinctive Identifiers

### DIFF
--- a/encrypted-media-respec.html
+++ b/encrypted-media-respec.html
@@ -1187,6 +1187,11 @@
                 </p>
               </li>
               <li>
+                <p>
+                  If the user agent requires informing the user for the <var>accumulated configuration</var> for other reasons, return <code>InformUser</code>.
+                </p>
+              </li>
+              <li>
                 <p>Return <code>Allowed</code>.</p>
               </li>
             </ol>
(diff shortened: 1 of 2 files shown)


mwatson2:
LGTM

PR #319, opened 2016-09-08
Fix #314: Informing the user and requiring user consent is always per origin

### LINKED ISSUE #314: Fix inconsistency: Informing the user and requiring user consent is always per-origin
When consent is required in the [Get Consent Status](https://w3c.github.io/encrypted-media/#get-consent-status) algorithm, the UA is to "Request user consent to use accumulated configuration **in the origin** and wait for the user response" (emphasis added).

When that algorithm returns `InformUser`, the UA is to "Inform the user that accumulated configuration is in use **in the origin**..."  (emphasis added).

https://w3c.github.io/encrypted-media/#security-prompts and https://w3c.github.io/encrypted-media/#privacy-prompts both say (emphasis added):
>Such alerts and consent **SHOULD** be per origin to avoid valid uses enabling subsequent malicious access and MUST be per browsing profile.

That "SHOULD" should be "MUST" to be consistent with the algorithms.

ddorwin:


mwatson2:
LGTM

mwatson2:
You forgot the index.html

ddorwin:
Thanks for catching that. Fixed.

PR #321, opened 2016-09-08
Fix #320: Clarify that consent is required unless all recommendations…

### LINKED ISSUE #320: Clarify that consent is required unless all recommendations for Distinctive Identifier clearing are implemented
The [Get Consent Status](https://w3c.github.io/encrypted-media/#get-consent-status) algorithm currently requires user consent if (among other conditions):
>The distinctiveIdentifier member of accumulated configuration is not "not-allowed" and the combination of the User Agent, implementation and accumulated configuration does not support the recommendations of Allow Persistent Data to Be Cleared with respect to Distinctive Identifier(s).

I believe the language should be tighter than "support the recommendations" for the following reasons:
 1. "Support" seems less definitive than implement or actually followed. (Below, I suggest "follow" since "implementation" is already used in the sentence.)
 1. We should be explicit that all recommendations must be followed, which is the intent.

Therefore, I propose replacing "support the recommendations" with "**follow all** the recommendations".

ddorwin:
… for Distinctive Identifier clearing are implemented

Replaces "support the recommendations" with "follow all the recommendations"

### DIFF
--- a/encrypted-media-respec.html
+++ b/encrypted-media-respec.html
@@ -1139,7 +1139,7 @@
                     <p>
                       The <a def-id="option-distinctiveIdentifier"></a> member of <var>accumulated configuration</var> is not
                       <a def-id="requirement-not-allowed"></a> and the combination of the User Agent, <var>implementation</var>
-                      and <var>accumulated configuration</var> does not support the recommendations of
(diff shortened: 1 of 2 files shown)


mwatson2:
LGTM

PR #322, opened 2016-09-09
Fix #315: Update headings for user informing and consent mitigation

### LINKED ISSUE #315: "Provide per-origin user alerts / prompts and permissions" headings do not accurately reflect the content
Both the Security and Privacy sections have mitigations with the heading "**Provide per-origin user alerts / prompts and permissions**".

The current heading could be interpreted as:
 1. The mitigation is to provide alerts or permission prompts.
 1. The mitigation is to ensure that any alerts, prompts, or permissions are per-origin.

More importantly, both contain important requirements that extend beyond alerts, prompts, or permissions. The per origin (and per browsing profile) requirements in the following paragraph are an additional mitigation. Certainly those requirements must (#314) be per-origin, but that is not the most important mitigation in these sections.

Specifically:
* https://w3c.github.io/encrypted-media/#security-prompts says, "User agents SHOULD ensure that users are fully informed and/or give explicit consent before a Key System that presents security concerns that are greater than other user agent features (e.g. DOM content) may be accessed by an origin."
* https://w3c.github.io/encrypted-media/#privacy-prompts says "User agents MUST ensure that users are fully informed and/or give explicit consent before using Distinctive Identifier(s) and Distinctive Permanent Identifier(s)."

ddorwin:
Also replace one reference to these sections with a more appropriate one.

### DIFF
--- a/encrypted-media-respec.html
+++ b/encrypted-media-respec.html
@@ -3737,7 +3737,7 @@ interface MediaEncryptedEvent : Event {
         <p>If a user agent chooses to support a Key System implementation that cannot be sufficiently sandboxed or otherwise secured, the user agent SHOULD ensure that users are fully informed and/or give explicit consent before loading or invoking it.
         </p>
         <p class="note">Granting permissions to unauthenticated origins is equivalent to granting the permissions to any origin in the presence of a network attacker.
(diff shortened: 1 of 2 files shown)


mwatson2:
LGTM

PR #323, opened 2016-09-09
Fix #316: Replace uses of "alerts"

### LINKED ISSUE #316: Replace uses of "alerts"
The spec currently uses "alert" seven times, all related to the "Provide per-origin user alerts / prompts and permissions" sections. Even within those sections, "alert" is not used for the main statement, which uses "ensure that users are fully informed" terminology. We should consistently use the latter, which more specifically addresses the requirement without implying a specific implementation.

Note that #315 may alter the structure of these sections.

ddorwin:


ddorwin:
The first commit is PR #322, so just review the second commit. The first will be a nop once that PR is merged.

encrypted-media-respec.html
```diff
@@ -3809,17 +3809,18 @@
         <section class="informative">
           <h4>Potential Attacks</h4>
           <p>Malicious pages could host legitimate applications in an iframe in an attempt hide an attack or deceive the user as to the source, such as making the use appear to be from a legitimate content provider.
-            This is especially relevent for user agents that support have <a href="#security-prompts">user alerts, prompts, and/or permissions</a>.
+            This is especially relevent for implementations that <a href="#security-consent">inform the user or require consent</a>, such as for security and/or <a href="#privacy-consent">privacy reasons</a>.
```
ddorwin: Originally, we only had one "prompts" section, but now we have two. This text applies to both, so I don't want to only refer to the one in the security section. I'm not particularly happy with this solution, but it does the job. Suggestions welcome! This also applies to line 3819.

encrypted-media-respec.html
```diff
@@ -3809,17 +3809,18 @@
...
-          <p>User agents are RECOMMENDED to base the UI and persistence of <a href="#security-prompts">user alerts, prompts, and/or permissions</a> on the combination of <a def-id="origin"></a> of the top-level <a def-id="document-concept"></a> and the <a def-id="origin"></a> using the APIs defined in this specification.
+          <p>User agents that <a href="#security-consent">inform the user or require consent</a>, including for security and/or <a href="#privacy-consent">privacy reasons</a>,
+            SHOULD base the UI and persistence of consent on the combination of <a def-id="origin"></a> of the top-level <a def-id="document-concept"></a> and the <a def-id="origin"></a> using the APIs defined in this specification.
             This ensures that users are informed of the main document making the request and that persisting a permission for one (legitimate) combination does not inadvertently allow malicious use to go undetected.
           </p>
-          <p>Authors are RECOMMENDED to prevent other entities from hosting their applications in <code>iframe</code>s.
+          <p>Authors SHOULD prevent other entities from hosting their applications in <code>iframe</code>s.
```
ddorwin: There is no change in requirements as this is a synonym. This just reads better. A similar change was made on line 3819.

mwatson2:
LGTM

PR #324, opened 2016-09-09
Remove explanation of why domain is insufficient

ddorwin:
The "above" reference is unclear but probably referred to the TLS section for
the Security instance. While the "above" reference could be replaced, this
statement is probably unnecessary since secure origins are required and origin
is explicitly used in the requirement.


mwatson2:
LGTM

PR #326, opened 2016-09-09
Move note that Distinctive Identifiers are not necessarily unique closer to the relevant condition

ddorwin:
Move only. No text changes.


mwatson2:
LGTM

PR #327, opened 2016-09-12
Several small commits related to value and identifier definitions

ddorwin:

PR #328, opened 2016-09-12
Fix #242: All values exposed to the application must be per-origin and clearable

### LINKED ISSUE #242: Privacy: Add statement that all values exposed or inferable by the application must be per-origin
We have a clear requirement that (most) [identifiers are per-origin](https://w3c.github.io/encrypted-media/#per-origin-identifiers) and that [persisted data must be clearable](https://w3c.github.io/encrypted-media/#allow-persistent-data-cleared), but I don't believe there is a single place that says most values (Permanent Identifiers being the obvious exception), both persisted and not, must be per-origin and not be accessible or inferable by other origins.

Some references to ~~#per-origin-identifiers~~ (now `#per-origin-per-profile-identifiers`) may not actually be related to identifiers and should instead point to such a requirement.

I believe there are some references in the `MediaKeys` and/or `MediaKeySessions` sections, and the web platform relies on origins, but it is still worth making a clear statement and having a reference target that is not specific to identifiers.

ddorwin:


ddorwin:
The first four commits are from #327, so just review the last commit. `index.html` includes changes from all five commits.

ddorwin:
There is now a sixth commit, so review everything _after_ the fourth commit. I will squash all of those into a single commit (the first four will be merged away).

encrypted-media-respec.html
```diff
@@ -3209,10 +3209,46 @@
...
+        </p>
+
+        <section id="per-origin-per-profile-values">
+          <h3>Use Per-Origin Per-Profile Values</h3>
+          <!-- Issue #101 may affect this text. -->
+          <p>
+            All <a href="#distinctive-value">distinctive values</a> exposed to or inferable by the application MUST be unique per <a def-id="origin"></a> and <a def-id="browsing-profile"></a>.
+            That is, the value(s) used for one <a def-id="origin"></a> using the APIs defined in this specification MUST be different from those used for any other origin using the APIs,
+            and value(s) used in one <a def-id="browsing-profile"></a> MUST be different from those used for any other profile, regardless of origin.
+          </p>
```
mwatson2: How does this apply to session ids ? Presumably if you pick session ids at random, then you could end up using the same one for different sessions in different origins.

ddorwin: Maybe we should replace "unique" with non-associable. Unique alone is insufficient if the values are `fixedID-origin`.

Such unintentional random collisions are okay (and preventing them could actually leak information in theory). We don't want intentional reuse, though.

Note that if IDs always start at 1 and increment, there would be similar collisions but these would not be distinctive.

ddorwin: Ah, non-associable is covered in the next paragraph.

As for your original point, maybe unique/different by design. However, do we really need to say anything? In theory, even a UUID can have collisions.

mwatson2: I think we can probable leave it, since there is not really a problem in practice, as you say.

My point is that all the issues related to distinctive values apply only if the value is provided to the application across multiple browsing sessions (allowing the application to say, "aha, I have seen this before, must be the same user"). A value could still be unique to a particular user, but if it local to a single browsing session, there's less of a tracking concern.

Now, with session ids, there is the possibility to query whether a given id is associated with a given user in a subsequent browsing session. This is a privacy issue only if the application already has another way to narrow down the set of possible users. If they have that, though, then an exhaustive search of possible session ids could reveal the user's identity.

Perhaps the definition of distinctive value should include that the value is visible across multiple browsing sessions ?

PR #330, opened 2016-09-16
Fix #329: Make MediaEncryptedEventInit required in the MediaEncryptedEvent constructor

### LINKED ISSUE #329: MediaKeyMessageEvent's message needs to be nullable or required in init dict
https://w3c.github.io/encrypted-media/#mediakeymessageevent

Per the current spec, `new MediaKeyMessageEvent('type')` would not throw an exception, because the init dict is not required. However, the result event's `message` attribute cannot sensibly be anything other than null, and yet it's not nullable in the IDL.

The typical pattern for event interfaces where only scripts could create instances with a certain member being null is to make that impossible, as such:

```WebIDL
[SecureContext,
 Constructor(DOMString type, MediaKeyMessageEventInit eventInitDict)]
interface MediaKeyMessageEvent : Event {
    readonly attribute MediaKeyMessageType messageType;
    readonly attribute ArrayBuffer         message;
};

[SecureContext]
dictionary MediaKeyMessageEventInit : EventInit {
    MediaKeyMessageType messageType = "license-request";
    required ArrayBuffer message;
};
```

If there is a concern for breaking existing content, the only other option is unfortunately to make `message` nullable.

ddorwin:
Also changes some init dictionary attributes to be required and not have default values.


ddorwin:
Virtually assigning to @foolip.

encrypted-media-respec.html
```diff
@@ -2332,8 +2332,8 @@
         <section>
           <h4><a>MediaKeyMessageEventInit</a></h4>
           <div><pre class="idl">[SecureContext] dictionary MediaKeyMessageEventInit : EventInit {
-             MediaKeyMessageType messageType = "license-request";
-             ArrayBuffer         message;
+             required MediaKeyMessageType messageType;
```
foolip: This is OK if you don't think there's any reasonable default. There's not a single hit for `MediaKeyMessageEventInit` in HTTP Archive, so I think legacy compat is not a problem.

encrypted-media-respec.html
```diff
@@ -2816,7 +2816,7 @@
         <section>
           <h4><a>MediaEncryptedEventInit</a></h4>
           <div><pre class="idl">dictionary MediaEncryptedEventInit : EventInit {
-             DOMString    initDataType = "";
+             required DOMString    initDataType
```
foolip: It seems a bit odd to require an `initDataType` if `initData` itself isn't required?

encrypted-media-respec.html
ddorwin: `"license-request"` is the most obvious one, but it's not really a "safe" default. It seems better to make the author specify it along with the message.

I don't expect anyone is actually using this directly.

encrypted-media-respec.html
ddorwin: I agree. I guess it wasn't clear what to do about the nullable initData, especially given the discussion in the issue.

Should this be `required ArrayBuffer? initData`?

Is the following valid in that case?
```
{ initDataType = "",
  initData = null; }
```

foolip: `initData` was already nullable on interface and init dict, so this seemed fine before, I think it was only because I assumed that the UA would never fire events with a null `initData` attribute that it seemed like a change to `MediaEncryptedEventInit` might be needed.

encrypted-media-respec.html
foolip: OK, just checking that there's *some* upside to a non-zero risk :)

foolip:
MediaKeyMessageEventInit changes LGTM

ddorwin:
This PR now only changes MediaEncryptedEvent & MediaEncryptedEventInit. I think it is ready to merge.

ddorwin:
s/MediaEncryptedEvent/MediaKeyMessageEvent/ above.

foolip:
Thanks @ddorwin, LGTM.

PR #331, opened 2016-09-16
Reorder the items in the User Tracking Mitigations section.

ddorwin:
Related items are now closer to each other.


ddorwin:
Addresses the TODO from PR #328 (and earlier).

Section moves only - no text changes other than removing the TODO.

The order may not be perfect, but it is much better. Suggestions welcome.

ddorwin:
No text change, so I'm going to merge.

PR #332, opened 2016-09-17
Fix #312: Require users be informed or provide consent when there are security concerns

### LINKED ISSUE #312: Security: Should implementations be required to inform users in cases where this is currently recommended?
The following are from the current Security section:

https://w3c.github.io/encrypted-media/#cdm-security:
>If a user agent chooses to support a Key System implementation that cannot be sufficiently sandboxed or otherwise secured, the user agent SHOULD ensure that users are fully informed and/or give explicit consent before loading or invoking it.

https://w3c.github.io/encrypted-media/#security-prompts:
>User agents SHOULD ensure that users are fully informed and/or give explicit consent before a Key System that presents security concerns that are greater than other user agent features (e.g. DOM content) may be accessed by an origin.

In cases where there are such concerns, should the spec _require_ (`MUST`) fully informing the user? This _might_ address some of the concerns in #304 and related discussion (i.e. in [this thread](https://lists.w3.org/Archives/Public/public-html-media/2016Sep/thread.html#msg3)).

ddorwin:


mwatson2:
I agree that the text changes in the PR accurately implement the proposal. So, if the proposal is agreed in principle, this PR does not require further review.

PR #342, opened 2016-10-24
Fix #333: Explicitly handle potentially-encrypted media data that does not contain Initialization Data

### LINKED ISSUE #333: Explicitly handle potentially-encrypted media data that does not contain Initialization Data
The spec currently handles "implementation [that require] specification of a MediaKeys object before decoding potentially-encrypted media data" in the [Initialization Data Encountered](https://w3c.github.io/encrypted-media/#initdata-encountered) algorithm. However, it is possible for a stream to be encrypted without it containing Initialization Data. Specifically, as pointed out in https://github.com/w3c/web-platform-tests/issues/3721#issuecomment-247526640, it is possible for CENC files to not contain `'pssh'` boxes.

The actual trigger for this step in CENC files is covered in the [Detection](http://w3c.github.io/encrypted-media/format-registry/stream/mp4.html#detect-format) section of the stream registry entry but not for this purpose.

I think we need to expand the registry entries to state that the Detection sections define encountering encrypted stream [data] then add a new algorithm for "Encrypted Stream Encountered" (or something similar) and move the step referenced in the first section to this algorithm. I don't think we need to duplicate it in the Initialization Data Encountered algorithm since I don't think Initialization Data should be encountered before it is determined that the stream is encrypted.

ddorwin:
Separates the relevant steps from the Initialization Data Encountered algorithm and specifies when to call them.

### DIFF
--- a/encrypted-media-respec.html
+++ b/encrypted-media-respec.html
@@ -2704,7 +2704,19 @@ interface MediaKeyMessageEvent : Event {
           <p>In addition to the criteria specified in [[!HTML5]], an <a def-id="htmlmediaelement"></a> SHALL be considered a <a def-id="blocked-media-element"></a> if its <var>waiting for key</var> value is <code>true</code>.</p>
         </li>
         <li>
+          <p>When the user agent is ready to begin playback and has encountered an indication that the <a def-id="media-data"></a> may contain encrypted blocks during the <a def-id="resource-fetch-algorithm"></a>, the user agent SHALL run the <a href="#media-may-contain-encrypted-blocks">Media Data May Contain Encrypted Blocks</a> algorithm.</p>
+          <p class="note">
+            For some container formats, such indication is separate from <a def-id="initialization-data"></a>.
+          </p>
+          <p class="note">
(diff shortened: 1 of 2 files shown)


ddorwin:
It should be possible for both "encrypted" events to be fired and for playback to be stopped. I believe this is how implementations behave. I've tried to specify this both normatively and in a note. Let me know if you have suggestions for improvement. We can also land this and iterate.

encrypted-media-respec.html
```diff
@@ -2859,6 +2871,26 @@
       <section id="htmlmediaelement-algorithms">
         <h3>Algorithms</h3>
 
+        <section id="media-may-contain-encrypted-blocks">
+          <h4>Media Data May Contain Encrypted Blocks</h4>
+          <p>
+            The Potentially Encrypted Media Encountered algorithm pauses playback if the user agent requires specification of a <a>MediaKeys</a> object before playing the media data.
```
mwatson2: Do you mean "The Media Data May Contain Encrypted Blocks Algorithm" here ?

PR #344, opened 2016-10-25
Fix #335: Note that many implementations only support decryption with MSE

### LINKED ISSUE #335: Spec should mention that EME is not expected to work outside of MSE
Currently the only browser that supports EME in non-MSE video is Chrome for WebM only. It doesn't seem like there's demand for EME in non-MSE video, so I propose that the spec mentions somewhere that EME is not expected to work outside of MSE video.

ddorwin:


ddorwin:
There was already a note to this effect in the `HTMLMediaElement` extensions section. This PR moves the note up to the introduction and directs it at authors.

@cpearce, please review.

PR #345, opened 2016-10-25
Fix #340: Make event preconditions consistent with HTML spec

### LINKED ISSUE #340: Preconditions in HTMLMediaElement Event Summary may not be consistent with HTML spec
The format of the [Event Summary](https://w3c.github.io/encrypted-media/#htmlmediaelement-events) table for HTMLMediaElement Extensions was adapted from the [Event Summary](https://html.spec.whatwg.org/multipage/embedded-content.html#mediaevents) in the HTML spec. Most of the Preconditions in that table appear to be a transition that has happened related to the cause of the event.

However, the Preconditions in the EME table are preconditions for getting to the point where the condition that causes them to be dispatched can occur.

ddorwin:
Also changes the Wait for Key to queue the waitingforkey event after setting readyState.
This is also more consistent with the HTML spec.

### DIFF
--- a/encrypted-media-respec.html
+++ b/encrypted-media-respec.html
@@ -2837,7 +2837,7 @@ interface MediaEncryptedEvent : Event {
               <td><a def-id="eventdfn">encrypted</a></td>
               <td><a>MediaEncryptedEvent</a></td>
               <td>The user agent encounters <a def-id="initialization-data"></a> in the <a def-id="media-data"></a>.</td>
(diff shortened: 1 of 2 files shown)


ddorwin:
This is based on top of PR #343, so ignore the first commit when reviewing.

The precondition for `waitingforkey` are now consistent with the HTML spec, including reordering the steps in the Wait for Key algorithm.

The precondition for `encrypted` didn't change except to copy the wording from the HTML spec. I did not see any additional text to add that was consistent with preconditions in the HTML spec.

ddorwin:
Rebased on top of the landed #343 to prepare for merging.

PR #348, opened 2016-10-26
Fix #347: Reset waiting for key and encrypted block queue when position changes

### LINKED ISSUE #347: Internal waiting for key and encrypted block queue values must be cleared on seek
The internal _waiting for key_ and _encrypted block queue_ values are initialized when the HTMLMediaElement is created:
>When a HTMLMediaElement is created, its internal _waiting for key_ value shall be initialized to false and its _encrypted block queue val_ue shall be empty.

But they are not cleared in other conditions. Seeking is the most obvious. Are there others?

@cpearce @foolip

ddorwin:


jdsmith3000:
This LGTM.

PR #349, opened 2016-10-26
Fix #336: Allow implementations to finish rendering frames before firing waitingforkey

### LINKED ISSUE #336: waitingforkey as spec'd could fire at different times in different browsers

ddorwin:


ddorwin:
This is an alternate fix for #336. It uses a variable to control whether the Wait for Key algorithm is run rather than trying to describe the condition as the previous PR #346 did.

jdsmith3000:
There's a fine difference between `decryption requires usable key` and `waiting for key`, and it feels like the former fills one of the reasons the latter was originally created.  It has a finer meaning though, indicating that decryption can't proceed, but not that the pipeline is stalled.  The existing `waiting for key` now only gets set once the pipeline can't proceed.  I agree the names don't make this very clear.

I wonder if we want names that directly state `decryption stalled waiting for key` and `playback stalled waiting for key`.

I like how this fix gets rid of the algorithm "continue" step.

ddorwin:
I'm fine with those names, though I worry that "stalled" has a different meaning the HTML spec.

@mwatson2, @cpearce, @foolip: Any comments on the naming or PR overall?

jdsmith3000:
Strictly speaking, only the decryption part is waiting for a key.  The pipeline is waiting for decrypted media.

encrypted-media-respec.html
```diff
@@ -3053,13 +3067,11 @@
...
-              <p class="note">These steps are reached when there is no key that is <a def-id="usable-for-decryption"></a> for <var>block</var>.</p>
-              <ol>
-                <li>
-                  <p>Run the <a def-id="wait-for-key-algorithm"></a> algorithm on the media element.</p>
-                </li>
-              </ol>
+              <p>Set the media element's <var>decryption requires usable key</var> value to <code>true</code>.</p>
+              <p class="note">This step is reached when there is no key that is <a def-id="usable-for-decryption"></a> for <var>block</var>.</p>
+              <div class="note">
+                <p>Once the user agent has rendered all the decrypted blocks that it can, it will run the <a def-id="wait-for-key-algorithm"></a> algorithm.</p>
```
mwatson2: I think this should be "...rendered all blocks preceding the block which cannot be decrypted, ...", or similar, since there may be unencrypted blocks mixed in.

encrypted-media-respec.html
```diff
@@ -2729,6 +2732,16 @@
           <p class="note">The above step provides flexibility for user agent implementations to perform decryption at any time after an encrypted block is encountered before it is needed for playback.</p>
```
mwatson2: We might want to clarify in the bullet above that we mean "encountered in the playback order" and that we're talking about a contiguous sequence of encrypted / unencrypted blocks. The corner case is that I have a sequence of decryptable / unencrypted blocks, followed by a missing block, followed by a non-decryptable block. We should wait for the missing block in the usual way (like a network stall) and only consider these algorithms with respect to the contiguous sequence when the missing block arrives and we can determine whether it is decryptable, unencrypted or undecryptable.

Maybe it is sufficient to say "This algorithm assumes blocks are encountered contiguously in playback order."

foolip:
I don't see "stalled" in the changes, but in HTML that's about the network, fired when no data has been delivered for 3 seconds IIRC.

ddorwin:
@foolip, "stalled" was in the suggested names in Jerry's comment. I used "blocked": _decryption blocked waiting for key_ and _playback blocked waiting for key_.

jdsmith3000:
"blocked" works for me.

ddorwin:
If there are no other comments, I'll merge this later today. We need to resolve the underlying issue, and I think we can address any remaining comments as separate issues or PRs.

PR #354, opened 2016-11-03
Fix #353: Remove persistent-usage-record from the V1 spec

### LINKED ISSUE #353: V1: Remove the at-risk persistent-usage-record feature and the related MediaKeySession destroyed algorithm, and remove the at risk notation for setting the media element readyState based on key availabity in two algorithms
These features are marked at-risk in the [July 5, 2016 EME CR document](http://www.w3.org/TR/2016/CR-encrypted-media-20160705/).  Based on the criteria applied by the working group for features to remain in the spec:

1.  _persistent-usage-record_ and the _MediaKeySystem destroyed algorithm_, with no currently passing implementations, shall be removed from the spec before it may advance.
2.  The _readyState_ behaviors may yet qualify as they are still being actively implemented and tested, and shall remain in the spec for the time being.
3.  At-risk notations for both (along with the at-risk section in general) shall be removed from the spec.

jdsmith3000:
Does not remove the at-risk section of the spec, but does attempt to remove all implementation details of persistent-usage-record.


jdsmith3000:
I neglected to do this PR in a branch, unfortunately.  Can someone advise me on how best to deal with that?

mwatson2:
@jdsmith3000 This is fine if we go with the `wicg/encrypted-media` fork option. You should also include the updated `index.html`.

jdsmith3000:
I think what I'd like to do is close this PR, save the changes, reset my gh-pages copy, create the branch and copy in the revised respec file.  Is that okay?

jdsmith3000:
I noticed the missing index as well.  I'm waiting to refresh it until I fix the PR, if needed.

jdsmith3000:
@mwatson2: Should I revert and close this PR, and do a new one off a branch?  If not, I'll add the index update.

mwatson2:
@jdsmith3000 If the proposal to fork to `wicg/encrypted-media` is agreed on the other thread, then this PR against `w3c/encrypted-media:gh-pages` is correct.

encrypted-media-respec.html
```diff
@@ -3616,9 +3431,7 @@
                 <li><p><a def-id="option-persistentState"></a>: Not <a def-id="requirement-required"></a> unless the application intends to create non-<a def-id="temporary-session"></a> sessions, if supported.</p></li>
               </ol>
             </li>
-            <li><p>The <a def-id="persistent-license-session"></a> <a>MediaKeySessionType</a>: Implementations MAY support this type.</p></li>
-            <li><p>The <a def-id="persistent-usage-record-session"></a> <a>MediaKeySessionType</a>: Implementations MAY support this type.</p></li>
-            
+            <li><p>The <a def-id="persistent-license-session"></a> <a>MediaKeySessionType</a>: Implementations MAY support this type.</p></li>
```
ddorwin: nit: There is new whitespace at the end of this line.

encrypted-media-respec.html
```diff
@@ -3756,23 +3563,13 @@
...
-            For sessions of type <a def-id="persistent-usage-record-session"></a> the object shall also contain the following members:
-          </p>
-          <dl>
-            <dt>"firstTime"</dt>
-            <dd>The <a def-id="first-decryption-time"></a> expressed as a number giving the time, in milliseconds since 01 January, 1970 UTC.</dd>
-            <dt>"latestTime"</dt>
-            <dd>The <a def-id="latest-decryption-time"></a> expressed as a number giving the time, in milliseconds since 01 January, 1970 UTC.</dd>
-          </dl>
-
+
```
ddorwin: nit: extra whitespace?

encrypted-media-respec.html
```diff
@@ -4051,7 +3834,7 @@
         <h3>User Tracking</h3>
         <section class="informative">
           <h4>Concerns</h4>
-          <p>A third-party host (or any entity, such as an advertiser, capable of getting content distributed to multiple sites) could use a <a def-id="distinctive-identifier"></a> or persistent data, including licenses, keys, key IDs, <a def-id="records-of-key-usage"></a>, or <a def-id="records-of-license-destruction"></a>, stored by or on behalf of the <a def-id="cdm"></a> to track a user across multiple sessions (including across <a def-id="origin">origins</a> and <a def-id="browsing-profile">browsing profiles</a>), building a profile of the user's activities or interests.
```
ddorwin: I missed fixing the order consistency here. In the fork, we should swap these for consistency. Alternatively, we can fix this then rebase this PR.

encrypted-media-respec.html
jdsmith3000: Fixed in next commit.

encrypted-media-respec.html
jdsmith3000: Fixed in next commit.

encrypted-media-respec.html
ddorwin: `<a def-id="records-of-key-usage"></a>, or <a def-id="records-of-license-destruction"></a>`
should be
`<a def-id="records-of-license-destruction"></a>, or <a def-id="records-of-key-usage"></a>`

I made the order consistent in a recent PR but missed this one.

ddorwin: Yes. I am about to push a fix. Then you can rebase.

ddorwin: Fixed in https://github.com/w3c/encrypted-media/commit/e56828dc3acfd56a6ceab874fa71c642a7816e72. You gan merge gh-pages in now.

jdsmith3000:
Changes have been rebased onto https://github.com/w3c/encrypted-media/commit/e56828dc3acfd56a6ceab874fa71c642a7816e72.

We've not opened the fork yet that preserves removed features.  Is it okay to merge the PR now?

I intend to do a separate PR/Merge to remove the at-risk features.

paulbrucecotton:
>Is it okay to merge the PR now?

No I think we have to fork first.

>I intend to do a separate PR/Merge to remove the at-risk feature section.

I believe you mean "to remove the list of at-risk features in the SOTD".

/paulc

jdsmith3000:
Correct.  This PR removes the persistent-usage-record details, but not the SOTD section on at-risk features.

jdsmith3000:
@ddorwin requested that at some point in case we wanted to revert the feature, but not restore the at-risk list.  I'm waiting to merge this PR before doing that last step.

ddorwin:
There's no reason we need to wait for the fork to land this. In fact, we'd want the fork to include the next commit @jdsmith3000 refers to anyway. Also, I think it will be easier to maintain (i.e. merge from upstream) if the fork contains a new commit rather than trying to maintain the lack of a commit. Thus, forking then reverting makes more sense.

To move forward with publishing PR (assuming the CfC passes), we should merge this PR and make the next change. Git makes it easy to branch from or recreate the state at any point in history, so we are not be blocked on creating a fork or branch.

jdsmith3000:
I agree with proceeding, and then reverting this commit in the fork.  If there are no objections (e.g. from @mwatson2), I'll do this today.

mwatson2:
Provided we are all agreed that the starting point in the master branch of WICG (or elsewhere, if that falls through for some reason) is our specification as it is today, including `persistent-usage-record`, I don't mind what order we do things in.

jdsmith3000:
That's essentially true.  The only difference will be that the list of at-risk features should be removed.

PR #356, opened 2016-11-04
Fix#353: Remove at-risk list from the SOTD.

jdsmith3000:


jdsmith3000:
I was expecting this PR to have just the SOTD change.  The previously merged commits are listed above as well.  Is this a problem?

ddorwin:
It looks like this is on your main branch. I'm not sure if that's a problem. But you need to rebase on upstream then `git push -f` to overwrite the branch. However, since this is on your main branch, you might just want to start over on a branch after merging upstream.

PR #359, opened 2016-11-08
Fix #358: Simplify wording about closing CDM session when MediaKeySession is destroyed

### LINKED ISSUE #358: Remove requirement to close CDM session *before* destroying UA MediaKeySession state
https://w3c.github.io/encrypted-media/#mediakeysession-interface says the following:
>If a `MediaKeySession` object becomes inaccessible to the page and is not closed, the User Agent MUST use the CDM to close the session **before User Agent state associated with the session is deleted**.

Is there a reason that the CDM session needs to be closed before deleting related UA state (the bold text)? There is no application-visible difference in when the CDM session is closed.

We ran into this while implementing this step. All CDM interactions are asynchronous and we're not going to block destruction of the `MediaKeySession`, so we will ignore the the bold text anyway. It seems like a superfluous requirement that we should just remove.

ddorwin:
Also use "key session" in several places to be consistent and differentiate.

### DIFF
--- a/encrypted-media-respec.html
+++ b/encrypted-media-respec.html
@@ -1596,7 +1596,7 @@
       </p>
       <p class="note">Closing the key session results in the destruction of any license(s) and key(s) that have not been explicitly stored.</p>
       <p class="note">
-        Exactly when the key session is closed is an implementation detail, and applications SHOULD NOT rely on this behavior.
+        Exactly when the key session is closed is an implementation detail, and applications SHOULD NOT rely on specific timing.
         <!-- TODO: Replace and/or expand on this with an implementation of Issue #360. --> 
         Applications that want to ensure a session is closed before taking some other action SHOULD call <a def-id="close"></a> and wait for the returned promise to be resolved.
       </p>
--- a/encrypted-media-respec.html
+++ b/encrypted-media-respec.html
@@ -1595,6 +1595,11 @@
         If a <a>MediaKeySession</a> object is not <a def-id="media-key-session-closed"></a> when it is destroyed, the CDM SHALL close the <a href="#key-session">key session</a> associated with the object.
       </p>
       <p class="note">Closing the key session results in the destruction of any license(s) and key(s) that have not been explicitly stored.</p>
+      <p class="note">
+        Applications SHOULD NOT rely on this behavior.
+        <!-- TODO: Replace and/or expand on this with an implementation of Issue #360. --> 
+        Applications that want to ensure a session is closed before taking some other action SHOULD call <a def-id="close"></a> and wait for the returned promise to be resolved.
+      </p>
 
       <p>For methods that return a promise, all errors are reported asynchronously by rejecting the returned Promise. This includes [[!WebIDL]] type mapping errors.</p>
       <p>The following steps of an algorithm are always aborted when rejecting a promise.</p>
--- a/encrypted-media-respec.html
+++ b/encrypted-media-respec.html
@@ -1402,7 +1402,7 @@
       <p>The MediaKeys object represents a set of keys that an associated HTMLMediaElement can use for decryption of <a def-id="media-data"></a> during playback.
         It also represents a CDM instance.
       </p>
-      <p>A MediaKeys object may be destroyed by the user agent when it is no longer accessible (i.e. there are no JavaScript references and no attached media element).</p>
+      <p>A <a>MediaKeys</a> object may be destroyed by the user agent when it is no longer accessible (i.e. there are no JavaScript references and no attached media element).</p>
       <p>For methods that return a promise, all errors are reported asynchronously by rejecting the returned Promise. This includes [[!WebIDL]] type mapping errors.</p>
       <p>The steps of an algorithm are always aborted when rejecting a promise.</p>
 
@@ -1584,7 +1584,7 @@
       <p>
         A <a>MediaKeySession</a> object SHALL NOT be destroyed and SHALL continue to receive events if it is
         not <a def-id="media-key-session-closed"></a> and the <a>MediaKeys</a> object that created it remains accessible.
-        Otherwise, a MediaKeySession object that is no longer accessible to JavaScript SHALL NOT receive further
+        Otherwise, a <a>MediaKeySession</a> object that is no longer accessible to JavaScript SHALL NOT receive further
         events and MAY be destroyed.
       </p>
       <p class="note">
--- a/encrypted-media-respec.html
+++ b/encrypted-media-respec.html
@@ -1592,11 +1592,11 @@
         objects and all <a>MediaKeySession</a> objects associated with the CDM instance are destroyed.
       </p>
       <p>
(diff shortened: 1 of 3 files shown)


encrypted-media-respec.html
```diff
@@ -1598,6 +1591,10 @@
         The above rule implies that the CDM instance must not be destroyed until all <a>MediaKeys</a>
         objects and all <a>MediaKeySession</a> objects associated with the CDM instance are destroyed.
       </p>
+      <p>
+        When a <a>MediaKeySession</a> object is destroyed, the CDM SHALL close the <a href="#key-session">key session</a> associated with the object.
```
mwatson2: The terminology 'when object X is destroyed' often gets interpreted in an implementation-dependent way, for example, the time at which garbage collection destroys some in-memory data structure. Should we use implementation-dependent language in our specification ?

The previous requirement was deliberately based on an implementation-independent condition.

ddorwin: I think during garbage collection is how this will be implemented. Is there a common concept of doing something when the "object becomes inaccessible?" If so, is that actually implemented in browsers? It seems that would require integration with the reference counting and garbage collection logic.

At least for the current steps, do we need to be more specific about the timing? In cases that applications actually care, they should call `close()` and wait for the returned promise. See the last part of https://github.com/w3c/encrypted-media/issues/358#issuecomment-259223670.

encrypted-media-respec.html
```diff
@@ -1581,15 +1581,8 @@
...
-        and is not <a def-id="media-key-session-closed"></a>, the User Agent
-        MUST use the CDM to close the session before User Agent state
-        associated with the <a href="#key-session">session</a> is deleted.
-      </p>
       <p>
-        A MediaKeySession object SHALL NOT be destroyed and SHALL continue to receive events if it is
+        A <a>MediaKeySession</a> object SHALL NOT be destroyed and SHALL continue to receive events if it is
         not <a def-id="media-key-session-closed"></a> and the <a>MediaKeys</a> object that created it remains accessible.
         Otherwise, a MediaKeySession object that is no longer accessible to JavaScript SHALL NOT receive further
         events and MAY be destroyed.
```
ddorwin: Interestingly, the spec does not require a MediaKeySession object to be destroyed when it is inaccessible. Maybe this is intended to allow GC to defer it (or for GC to never be activated if it is not necessary).

encrypted-media-respec.html
mwatson2: The reason to use "object becoming inaccessible" is that this is an event that is unambiguous across browsers. There is a point in time at which the last reference from Javascript to the object is dropped and this is the same point in time for every implementation, because there is a specific Javascript instruction in the page that causes the last reference to be dropped.

So, the point is to say that as a consequence of this happening, the CDM session will be closed.

The previous text did say "if ...". This is a straightforward logical requirement that doesn't specify any timing (so it does not imply "when"). But it does imply that closure of the session happens _after_ the object becomes inaccessible because effects always come after causes :-)

ddorwin:
The latest commit restores the missing check and uses slightly less prescriptive wording about timing.

encrypted-media-respec.html
ddorwin: I don't think that would accurately reflect any implementation, though. Nor do I think authors should rely on this. I've added a non-normative note to this effect.

Garbage collection seems to be the best platform concept we have for such things. For example, see https://html.spec.whatwg.org/multipage/embedded-content.html#best-practices-for-authors-using-media-elements. (We could replace "destroyed" with "garbage collection occurs") Also note that it is recommended that authors explicitly call `load()` rather than waiting for garbage collection.

Other than potentially replacing "destroyed," I believe the current commit (a60a130) is consistent with how HTML5 deals with similar resource release issues.

mwatson2:
I think that as worded this is a substantive change.

Previously, applications could rely on the CDM session being closed (and the licenses and keys being destroyed) if the MediaKeySession became inaccessible. They could not rely on the _timing_ of this (except that it must be before all record of the key session was expunged), but the specification required that it happened.

With the proposed change, this is only required when the MediaKeySession is "destroyed" and it is unspecified both when _and whether_ this would happen (or what "destroyed" means at all).

I think we should retain the previous text that specified that closure of the session must happen, without prescribing exactly when.

ddorwin:
>I think that as worded this is a substantive change.

Without commenting on this specific statement, I'm not sure that we are prohibited from making such a change until the PR has been produced and the transition request made. We would definitely want to avoid substantive changes after that point.

>Previously, applications could rely on the CDM session being closed (and the licenses and keys being destroyed) if the MediaKeySession became inaccessible. They could not rely on the _timing_ of this (except that it must be before all record of the key session was expunged), but the specification required that it happened.

I don't think application could rely on that, at least not in a robust way. Only the CDM and license policy can enforce restrictions. (As has been discussed, there are mechanisms for limiting the extent of the policy and/or recording that this actually happened, but there was never a guarantee that it would happen.

>With the proposed change, this is only required when the MediaKeySession is "destroyed" and it is unspecified both when _and whether_ this would happen (or what "destroyed" means at all).

>I think we should retain the previous text that specified that closure of the session must happen, without prescribing exactly when.

I think the previous text is incorrect and/or misleading as I believe it implies more specific timing than we intend.

I am fine with saying the "closure of the session **_should_** happen, without prescribing exactly when." (I say "should" because, as I mentioned above, there are no guarantees about what will happen in the UA. This is likely moot since we wouldn't actually use this language.) Thus, the question is how do we say something should happen without specifying when?

paulbrucecotton:
>Without commenting on this specific statement, I'm not sure that we are prohibited from making such a change until the PR has been produced and the transition request made. We would definitely want to avoid substantive changes after that point.

As Chair I have a STRONG preference for not making "substantive changes" during the WG CfC since that could force us to restart that clock.

ddorwin:
The latest commit restores the "becomes inaccessible to the page" language without implying specific timing.

encrypted-media-respec.html
```diff
@@ -1581,23 +1581,25 @@
...
       <p class="note">
         The above rule implies that the CDM instance must not be destroyed until all <a>MediaKeys</a>
         objects and all <a>MediaKeySession</a> objects associated with the CDM instance are destroyed.
       </p>
+      <p>
+        If a <a>MediaKeySession</a> object is not <a def-id="media-key-session-closed"></a> when it becomes inaccessible to the page, the CDM SHALL close the <a href="#key-session">key session</a> associated with the object.
+      </p>
+      <p class="note">Closing the key session results in the destruction of any license(s) and key(s) that have not been explicitly stored.</p>
+      <p class="note">
+        Exactly when the key session is closed is an implementation detail, and applications SHOULD NOT rely on this behavior.
```
mwatson2: s/this behavior/specific timing/ perhaps ?

Previously, there was a specification requirement that this should happen (without saying when). A specification requirement is not a guarantee, but it is a requirement.

mwatson2:
With David's last change and a small clarification [I proposed](https://github.com/w3c/encrypted-media/pull/359#discussion_r87114926), this becomes a useful clarification rather than a substantive change (IMO).

ddorwin:
I made Mark's suggested change. Please confirm this is ready to merge.

PR #367, opened 2017-01-10
Fix length of pssh box in example

tedjp:
The example pssh box is listed as having a length of 76, but the entire structure is actually 68 bytes.


tedjp:
Marked as non-substantive for IPR from ash-nazg.

plehegar:
@ddorwin any reason not to merge this?

ddorwin:
The `cenc-respec.html` file needs to be updated too. If ReSpec works on these files, we can generate the `cenc.html` file. If not, we should update the date too.

plehegar:
oh, I deleted the gh-pages branch but didn't realize it would close this issue. I'll send a separate pull request to address this.

plehegar:
#399 should provide the proper fix

PR #392, opened 2017-05-02
Add ISO Common Encryption 'cbcs' Protection Scheme. Fixes #391.

### LINKED ISSUE #391: Add support for ISO Common Encryption 'cbcs' Protection Scheme
See the pull request https://github.com/w3c/encrypted-media/pull/392 for details.

hober:


hober:
I don't know why the IPR check failed. I am in the HME WG.

paulbrucecotton:
The HME WG's charter expired on Apr 30 so until the charter is extended no one is actually a member of the HME WG.  ;<)

@plehegar needs to fix this.

/paulc

plehegar:
Adding a protection scheme would probably impact the tests, right? If yes, then this seems a late request imho.

hober:
This change brings the spec more into alignment with shipping implementations. For instance, Safari has supported cbcs since it first shipped EME. The spec should reflect reality here.

mwatson2:
@hober We weren't able to run the test suite with Safari back last year. Have things changed and Safari is now spec-compliant ?

jernoble:
@mwatson2 WebKit is working on improving our compatibility with the test suite. But apart from bringing our DOM APIs up to spec, we have the additional limitation on macOS that our platform only supports the 'cbcs' scheme. Hence our desire to document Safari's shipping behavior.

hober:
What's the rationale for deferring this to v.next?

paulbrucecotton:
>What's the rationale for deferring this to v.next?

EME V1 is done and the Proposed Recommendation is being dealt with by the Director and Advisory Committee.  IFF EME becomes a Recommendation the current [HME WG charter](https://lists.w3.org/Archives/Public/public-html-media/2017May/0008.html) would only permit us to work on non-breaking changes errata.

/paulc
HME WG Chair

hober:
Fair enough. Any reason to not spin up an Editor's Draft of small changes like this one, so that we can hit the ground running when the way forward is clearer?

paulbrucecotton:
Note that this pull request actually proposes changes to the non-normative [Encrypted Media Extensions Stream Format Registry ](http://www.w3.org/TR/eme-stream-registry/) which was last published as a WG Note in Sep 2016.  I believe this kind of change can be made at any time IFF we can get consensus on the change.

/paulc
HME WG Chair

joeyparrish:
Any objections to merging this now?  I'll be sending a PR soon for an API change to query encryption scheme support, including 'cbcs', and this seems like a needed compliment to that.

hober:
@joeyparrish I'm all for it.

joeyparrish:
I'll go ahead and merge it, then.  I may make a few edits on top of that as I work to complete PR #457.

hober:
> I'll go ahead and merge it, then.

Thanks!

Issue #383, opened 2017-03-31
Sandboxing expectation


plehegar:
Section 10.2 says:
[[
User Agents are responsible for providing users with a secure way to browse the web, including any functionality, such as CDMs, from third parties
[...]
Exploiting a CDM implementation that is not fully sandboxed and/or uses platform features may allow an attacker to access OS or platform features [...]
If a user agent chooses to support a Key System implementation that cannot be sufficiently sandboxed or otherwise secured, the user agent SHOULD ensure that users are fully informed and/or give explicit consent before loading or invoking it. 
]]

However, this expectation isn't really reflected earlier in the specification, especially the graph in Section 1 Introduction.
#382 proposes to address this issue by adding a sandbox around the CDM in the graph.

dwsinger:
I can't even parse (from an english point of view) the first sentence "User Agents are responsible for providing users with a secure way to browse the web, including any functionality, such as CDMs, from third parties.".  What is 'including' attached to? 'the web'? 'a secure way'? but CDMs are not 'from the web'... can someone re-write this sentence to make it closer to comprehensible by mere motals?

plehegar:
@mwatson2 @ddorwin any chance one of you could look at this?

plehegar:
How about "User Agents are responsible for providing users with a secure way to browse the web. This responsibility covers any functionality used by User Agents, including functionalities from third parties like CDMs."

mwatson2:
Is this to be an Editorial in the REC, or for future work ?

Your proposal is what I believe was intended by that text, yes.

mavgit:
Given that:
a. there are many user agents that are written by one party as applications on top of operating systems written by another party, and
b. there are a great many key web APIs which are interfaces to functionality that is often in large part implemented by operating systems as a service to applications, such as File API, indexedDB, XHR, Geolocation, Canvas, WebGL, KeyboardEvent, TouchEvent, MSE and, in some OSs, EME.

While there are security expectations of browser applications and also security expectations of OSs, I don't think that there are (or could reasonably be) expectations that a browser application is responsible for the security of OS services implemented by another party, when the browser accesses those services by one of the APIs above. For example, I don't think anyone expects that a Chrome browser downloaded to Windows becomes responsible for Windows file system security, just because the Chrome browser accesses the Windows file system via the File API.

I think this applies to all the APIs above (and other OS-interface APIs), including EME.

Thanks,
mav

[...]

PR #394, opened 2017-05-10
Emphasized security and privacy in the status

plehegar:
@mwatson2 @ddorwin This is an attempt to be more upfront in the specification about its intent regarding security and privacy. The proposed note is added in the status of the document.

### DIFF
--- a/encrypted-media-respec.html
+++ b/encrypted-media-respec.html
@@ -159,6 +159,12 @@
 
     <section id="sotd">
         <p>The <code>"persistent-usage-record"</code> session type and the related <em>MediaKeySession destroyed</em> algorithm were removed since the previous version.</p>
+        <p class="note">
+          Implementers are advised to pay extra attention to sections <a href='#security'></a> and <a href='#privacy'></a>, to mitigate threats and vulnerabilities from
+          underlying Key System implementations, and to provide users with adequate control over their own privacy. If a user agent chooses to support a
+          Key System implementation that cannot be sufficiently sandboxed or otherwise secured, the user agent SHOULD ensure that users are fully informed and/or give explicit
+          consent before loading or invoking it.
+        </p>
     </section>
 
     <section id="introduction" class="informative">
--- a/encrypted-media-respec.html
+++ b/encrypted-media-respec.html
@@ -159,15 +159,6 @@
 
     <section id="sotd">
         <p>The <code>"persistent-usage-record"</code> session type and the related <em>MediaKeySession destroyed</em> algorithm were removed since the previous version.</p>
-        <p class="note">
-          Implementers are advised to pay extra attention to sections <a href='#security'></a> and <a
-          href='#privacy'></a>, to mitigate threats and vulnerabilities from underlying Key System
-          implementations, and to provide users with adequate control over their own privacy. In particular,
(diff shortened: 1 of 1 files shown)


mwatson2:
@plehegar I am fine with this text, but we could go further. I think it is an implication of the first sentence that User Agents cannot support Key Systems with unknown (to them) privacy or security properties. Otherwise, how can they guarantee compliance to the privacy or security sections ? But perhaps this could be called out more explicitly. For example:

> In particular, the requirements of the privacy and security sections cannot be met without knowledge of the privacy and security properties of the Key System.

plehegar:
Thank you. I added your suggestion in the paragraph.

plehegar:
This would address https://github.com/w3c/encrypted-media/issues/383 imho

jdsmith3000:
This change is fine with me.  It appears ready to merge.

encrypted-media-respec.html
```diff
@@ -159,6 +159,15 @@
 
     <section id="sotd">
         <p>The <code>"persistent-usage-record"</code> session type and the related <em>MediaKeySession destroyed</em> algorithm were removed since the previous version.</p>
+        <p class="note">
+          Implementers are advised to pay extra attention to sections <a href='#security'></a> and <a
```
ddorwin: Should it be "...to **the** foo and bar **sections**"?

encrypted-media-respec.html
```diff
@@ -159,6 +159,15 @@
...
+          href='#privacy'></a>, to mitigate threats and vulnerabilities from underlying Key System
+          implementations, and to provide users with adequate control over their own privacy. In particular,
+          the requirements of the privacy and security sections cannot be met without knowledge of the privacy
+          and security properties of the Key System. If a user agent chooses to support a Key System
```
ddorwin: Add "implementation" after Key System. Or "Key System and its implementation(s)".

encrypted-media-respec.html
```diff
@@ -159,6 +159,15 @@
 
     <section id="sotd">
         <p>The <code>"persistent-usage-record"</code> session type and the related <em>MediaKeySession destroyed</em> algorithm were removed since the previous version.</p>
```
ddorwin: Unrelated: Should this be removed since it no longer applies to the next time the spec is published? (We already published a version (PR) with this text.)

ddorwin:
@plehegar wrote:
>This would address #383 imho

A concern about some existing text was raised in https://github.com/w3c/encrypted-media/issues/383#issuecomment-290893383, and we should also address that. I made a proposal in https://github.com/w3c/encrypted-media/issues/383#issuecomment-293104384, but there may be other options.

ddorwin:
I filed #405 to track the concern I mentioned above, so we can ignore my previous comment.

ddorwin:
Is the SotD the right section for this statement? This will always be true of the document and is not specific to the current status of the document.

jdsmith3000:
I agree this paragraph feels out of place in the SofD.  It's not status (as @ddorwin pointed out) and also talks about Key Systems before introducing them.  It might work in the Introduction, but probably needs a preface on Key Systems in EME.

encrypted-media-respec.html
ddorwin: Addressed in #410.

PR #395, opened 2017-05-10
Fix an incorrect markup correction

plehegar:
https://github.com/w3c/encrypted-media/commit/1f112baec28a70b4c472902c57d9856fb552288b#diff-f72607e47a6f74e53dc90eab8ee094e2 made the NOTE apply to only one branch of the step. If the previous markup was incorrect, the note should have been moved outside the `</dl>`.


jdsmith3000:
Change is editorial and LGTM.

PR #396, opened 2017-05-10
Undoing improper fragment fixes

plehegar:
The last two commits changes macros, especially those related to queueing a task. There are a couple issues.

    The `queue-a-task` macro name was replaced with `queueing`. I think this should have been left alone as it more clearly represents what the macro will be converted to. It doesn't have to match the target anchor name.
    Upper case 'Q' was replaced with 'q' in the macro names. The upper case letter is intentional because it indicates that he text will be capitalized. For one macro, we had both upper and lower case 'Q'. Now both are lower case, which breaks some of the resulting text. For example, step 10.10 of https://w3c.github.io/encrypted-media/#dom-mediakeysession-generaterequest.

### DIFF
--- a/encrypted-media-respec.html
+++ b/encrypted-media-respec.html
@@ -1538,7 +1538,7 @@
           <p>The following step is run:</p>
           <ol>
             <li>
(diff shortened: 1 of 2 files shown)


jdsmith3000:
Change looks complete to me.  Ready to close.

PR #398, opened 2017-05-10
Fix #393: encrypted event is of type encrypted

### LINKED ISSUE #393: Typo in "7.3.2 Initialization Data Encountered"
Step 5 says "with its type attribute set to message" while it should be "with its type attribute set to encrypted".

plehegar:
Fixes
 https://github.com/w3c/encrypted-media/issues/393


ddorwin:
Also, when merging, please use the updated subject rather than the default one that will be populated.

PR #399, opened 2017-05-10
Fix #367: Correct length of pssh box in example 

plehegar:
Replaces #367


ddorwin:
LGTM. Should we update the date on this page?

ddorwin:
@plehegar We should update the date, right?

PR #401, opened 2017-05-18
Fix #372: setMediaKeys() step 1 and 2 should be reversed

### LINKED ISSUE #372: setMediaKeys() step 1 and 2 should be reversed
The first two steps of [setMediaKeys()](https://w3c.github.io/encrypted-media/#dom-htmlmediaelement-setmediakeys) are as follows:

> 1. If mediaKeys and the mediaKeys attribute are the same object, return a resolved promise.
> 2. If this object's attaching media keys value is true, return a promise rejected with an InvalidStateError.

Now imagine the application makes the following calls:
```
video.setMediaKeys(mediaKeys);
video.setMediaKeys(null);
```
On the second call, the promise of the first call could still be pending, e.g. the result is in a task waiting in a queue. So the _mediaKeys attribute_ is still null. Then according to step 1, since both the _mediaKeys_ and _mediaKeys attribute_ are null, in theory they **are** the same object, a resolved promise is returned. Then, when the task in the queue runs, the promise of the first call is also resolved.

This could confuse the application that even if the promise of the second call is resolved, the _mediaKeys attribute_ is not null.

Should we switch the order of step 1 and step 2, such that we always check the _attaching media keys value_ first, and any setMediaKeys() during another pending setMediaKeys() call will always be rejected?

ddorwin:


jdsmith3000:
LGTM, pending a decision whether it can be merged for V1.

plehegar:
Approved by editors.

PR #402, opened 2017-05-18
Fix #363: setServerCertificate() should sanitize the input before providing it to the CDM

### LINKED ISSUE #363: setServerCertificate() should sanitize the input before providing it to the CDM
Data provided from the application to the CDM is sanitized in all of the `MediaKeySession` methods, but the input to [`setServerCertificate()`](https://w3c.github.io/encrypted-media/#setServerCertificate) is not sanitized. The spec should be consistent in sanitizing inputs.

ddorwin:


jdsmith3000:
LGTM as well.

PR #404, opened 2017-05-18
Fix #403: Remove [SecureContext] from enum and dictionary definitions

### LINKED ISSUE #403: Remove [SecureContext] from enum and dictionary definitions
My reading of https://heycam.github.io/webidl/#SecureContext is that `[SecureContext]` is not allowed on `enum` and `dictionary` definitions. Thus, we should remove this non-compliant Web IDL.

This was first raised in https://github.com/w3c/encrypted-media/issues/289#issuecomment-298016695.

ddorwin:


jdsmith3000:
Looks good to me.

PR #411, opened 2017-06-08
Fix #407 and #408: Add CDM Constraints and move Definitions text

jdsmith3000:
This PR adds overview text on CDM constraints and relocates CDM messaging requirements to section **8. Implementation Requirements**.  These changes were discussed in issues 307 & 308.

***
[Preview](https://s3.amazonaws.com/pr-preview/jdsmith3000/encrypted-media/cdm-constraints-definition-fixes.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/encrypted-media/2aa65a9...jdsmith3000:fba8986.html)

PR #412, opened 2017-06-09
Address post-merge comments from PR #411

ddorwin:


encrypted-media-respec.html
```diff
@@ -3023,17 +3027,17 @@
         <ul>
           <li>
             <p>
-              Access network resources, either local or remote, except via the user agent, or as part of the user agent, as explicitly permitted by this specification.
+              Access network resources, either local or remote, except via the user agent as explicitly permitted by this specification.
```
mwatson2: @jdsmith3000 had commented that the CDM being "part of the user agent" was an explicitly allowed implementation choice  (per the paragraph moved back to the CDM definition above). I understood this to be the reason for the "or as part of the user agent" text. However, this case is IMO a subset of "via the user agent" in the sense that a department of FedEx can still send something via FedEx.

ddorwin: I agree that such implementations are allowed and that "via the user agent" includes this.

As I said in #411, the CDM is always _logically_ distinct from the user agent regardless of how they are implemented, so I don't think this is necessary. Since it increases the complexity and decreases the clarity of what is intended to be a clarifying statement, I recommend we remove it.

jdsmith3000: Most changes in this PR look fine, but I disagree with removing "as part of the user agent".  That text is necessary for completeness.  The goal of this section is to provide an overview of the implementation requirements for security and privacy, specifically where sandboxing is one implementation approach.  The implementation note (which this PR moves back to the CDM definition) is the sole area where this option is stated, and if moved will be remote from this overview paragraph.  I believe moving it back by itself is acceptable, but the "as part of the user agent" text really be maintained for completeness.  Without it, this implementation option may be clear to members of the working group, but it's hardly clear to people reading the spec for the first time.

ddorwin: I added a statement to the parent section as discussed offline.

jdsmith3000:
LGTM with last change.

PR #414, opened 2017-06-09
Refer to Implementation Requirements, Security, and Privacy sections implicitly

ddorwin:
This change causes these these sections to automatically be referenced by section number and title.


ddorwin:
This is just a proposal. We do this in some other places in the spec.

Issue #369, opened 2017-01-24
Unclear text about application managing storage generated by CDM


xhwang-chromium:
In the current spec on [persistent-license](https://w3c.github.io/encrypted-media/#dom-mediakeysessiontype-persistent-license), it says:

> The session must be loadable via its Session ID once update() is called successfully. **The application is responsible for managing any such storage that may be generated by the CDM.** See Session Storage and Persistence.

It's not clear to me what "such storage" generated by the CDM the application is responsible for managing. After some offline discussion (with ddorwin@), this is really about the text appeared at the end of [6.6 Session Storage and Persistence](https://w3c.github.io/encrypted-media/#session-storage):

> An application that creates a session using a persistent-license should later remove the stored data using remove(). The CDM may also remove sessions as appropriate, but applications should not rely on this.

However, it's pretty hard to figure this out from the current text. We should clarify this part to avoid any future confusion.

mwatson2:
The text in 6.6 is clear (if #368 is addressed) so perhaps we should just remove the first phrase mentioned above ?

jdsmith3000:
The wording under persistent-license sessions seems like it's intended to introduce the section 6.6 content.  I don't think it's improved by removing it.  I recommend leave as is.

mwatson2:
Perhaps s/managing any such storage that may be generated/requesting the removal of session data stored/ ?

jdsmith3000:
Works for me.

ddorwin:
"any such storage" is referring to the things in the first paragraph:
>A session for which the license (and potentially other data related to the session) will be persisted. A record of license destruction shall be persisted when the license and key(s) it contains are destroyed.

The sentence containing "any such storage" and the next one seem out of place in the middle of a paragraph describing client implementation requirements. Also, the middle and last paragraph are out of order with respect to lifetime.

We have an incomplete sentence beginning with "Can only." I think we should change this to "Such sessions can only..."

Something like @mwatson2's suggestion seems reasonable. I'll make a PR to address the comments above, then we can see what makes sense to address this.

ddorwin:
See PR #417 for structural changes.

As for addressing the original issue, I suggest something similar to one of the following:

>The application is responsible for ensuring that data persisted for such sessions is removed when the application no longer needs it. See Session Storage and Persistence.

>The application is responsible for removing the session, and thus data persisted for it, when the application no longer needs it. See Session Storage and Persistence.

xhwang-chromium:
[...]

PR #417, opened 2017-06-14
Minor reorganization of the persistent-license definition

ddorwin:
Per the discussion in issue #369.


ddorwin:
This addresses the structure comments in https://github.com/w3c/encrypted-media/issues/369#issuecomment-308519868. The changes are only moves except for "Sessions of this type can..."

PR #419, opened 2017-06-14
Use "this object" rather than a "session" variable in close()

ddorwin:
This is consistent with the other MediaKeySession method algorithms.


ddorwin:
The current text may be left over from refactoring related to the Session Closed algorithm, which uses a _session_ variable.

ddorwin:
This is fairly mechanical and blocking another PR, so I'm going to merge it. We can address feedback separately.

PR #421, opened 2017-06-14
Fix #365: Do not allow operations after close() is called but before the session is actually closed

### LINKED ISSUE #365: Operations are allowed after close() is called but before the session is actually closed
[MediaKeySession.close()](http://w3c.github.io/encrypted-media/#dom-mediakeysession-close) specifies:
step 2: If session is closed, return a resolved promise.
step 5.3: Queue a task to run the following steps:
step 5.3.1: Run the Session Closed algorithm on the session.

However, the definition of ["closed"](http://w3c.github.io/encrypted-media/#media-key-session-closed) from step 2 states "MediaKeySession object is closed if and only if the Session Closed algorithm has been run." Depending upon timing, it is possible that an application could call close() multiple times before the Session Closed algorithm has actually run.

This would apply to any operation that checks "if the session is closed ..." before running steps in a separate task.

ddorwin:
Introduces a new "closed" value and use it when determining whether to abort MediaKeySession algorithms.

Also, make media-key-session-closed depend on an explicit state.
Previously, this definition depended on whether an algorithm had been run, which is slightly ambiguous.

### DIFF
--- a/encrypted-media-respec.html
+++ b/encrypted-media-respec.html
@@ -2355,11 +2355,13 @@ interface MediaKeyMessageEvent : Event {
           <p>The following steps are run:</p>
           <ol>
             <li><p>Let <var>session</var> be the associated <a>MediaKeySession</a> object.</p></li>
+            <li><p>Let <var>promise</var> be the <var>session</var>'s <a def-id="closed"></a> attribute.</p></li>
+            <!-- Avoid a race condition that could cause the algorithms below to be run multiple times. -->
+            <li><p>If <var>promise</var> is resolved, abort these steps.</p></li>
             <!-- Handle the case that this algorithm is reached by a path other than close(). In all cases, the value is set before the algorithms below are run. -->
-            <li><p>Set this <var>session</var>'s <var>closing or closed</var> value to true.</p></li>
+            <li><p>Set the <var>session</var>'s <var>closing or closed</var> value to true.</p></li>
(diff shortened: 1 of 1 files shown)


ddorwin:
The first commit fixes the race condition identified in #365. We might want to rename the value, though since this value's state does not match the definition of "closed," which is defined at `#media-key-session-closed`. There is also a `closed` attribute. Because of the various paths to setting the new closed value, I'm having trouble finding a better name. It is really "closed to new calls" or "closed or pending close." Suggestions welcome, or we can merge this and rename it later.

The second commit is minor cleanup, which should make other algorithms more deterministic and unambiguous. It is unrelated to the race condition in #365.

mwatson2:
This LGTM. Should there be a check at the top of the Session Closed algorithm that checks whether the `closed` promise is already resolved ? This would handle the case where two tasks to run Session Closed are queued (one by the Monitor CDM algorithm and one by `close()`).

Perhaps the new variable should be called _closing_ ? Or to be strictly correct  _closing-or-closed_ ?

ddorwin:
I made the changes @mwatson2 suggested.

mwatson2:
Looks good. Should _closing or closed_ be hyphenated, though (_closing-or-closed_). I'm concerned that the italicization of "or" is the only thing that indicates this is one variable not two (_closing_ or _closed_).

ddorwin:
I agree that would be clearer, but that would be inconsistent with all other such values (search for "value be" in the spec). This is really a style problem and should maybe be addressed with CSS.

PR #423, opened 2017-06-15
Improve readability of the text defining key ID comparison

ddorwin:


ddorwin:
The existing sentence on sorting is very long with lots of fragments. This change makes sorting and  comparison separate sentences and tries to make the latter sentence clearer. I also swapped `m` and `n` for alphabetical consistency.

For reference, this text was originally added in https://github.com/w3c/encrypted-media/commit/76c54ae413d81492d0823a482ab41dbbc389afe7.

PR #424, opened 2017-06-15
Fix #370: Resolve ambiguous explanations of time

### LINKED ISSUE #370: Resolve ambiguous and/or contradictory explanations of time
As discussed in https://github.com/w3c/encrypted-media/issues/59#issuecomment-275488967 and subsuquent comments, the definition of [Time](https://w3c.github.io/encrypted-media/#time) is ambiguous and/or contradictory. The same is true of references to time elsewhere in the spec.

Specific changes include:
1. Change all references to the unofficial ES6 draft to the official spec. Specifically, replace https://people.mozilla.org/~jorendorff/es6-draft.html#sec-date-objects with https://tc39.github.io/ecma262/#sec-time-values-and-time-range.
1. Rely on https://tc39.github.io/ecma262/#sec-time-values-and-time-range, providing only clarifying text when necessary. Perhaps, highlight the important parts in a non-normative Note to avoid redundantly and potentially confusingly defining time.
1. Ensure references to "Unix time" are removed.
1. Be explicit that time is represented in milliseconds ("millisecond accuracy" and "instants... it can represent... every millisecond" are not sufficiently explicit).
1. In algorithms, reference the Time definition rather than defining the meaning of the time value in each algorithm. For example, "...let expiration time be the [new] expiration time in milliseconds since 01 January 1970 UTC" is redundant and incomplete.

jdsmith3000:


encrypted-media-respec.html
```diff
@@ -1811,7 +1810,7 @@
                       <p class="note">In other words, do not create a session if a non-closed persistent session already exists for this <var>sanitized session ID</var> in any browsing context.</p>
                     </li>
                     <li><p>Load the <var>session data</var>.</p></li>
-                    <li><p>If the <var>session data</var> indicates an expiration time for the session, let <var>expiration time</var> be the expiration time in milliseconds since 01 January 1970 UTC.</p></li>
+                    <li><p>If the <var>session data</var> indicates an expiration time for the session, let that be the <var>expiration time</var>.</p></li>
```
mwatson2: `that` is not all that clear. Perhaps

> If the _session data_ indicates an expiration time for the session, let _expiration time_ be that expiration time.

encrypted-media-respec.html
```diff
@@ -2326,7 +2325,7 @@
           <ol>
             <li><p>Let the <var>session</var> be the associated <a>MediaKeySession</a> object.</p></li>
             <li><p>Let <var>expiration time</var> be <code>NaN</code>.</p></li>
-            <li><p>If the new expiration time is not <code>NaN</code>, let <var>expiration time</var> be the new expiration time in milliseconds since 01 January 1970 UTC.</p></li>
+            <li><p>If the new expiration time is not <code>NaN</code>, let that be a new <var>expiration time</var>.</p></li>
```
mwatson2: Ditto. Perhaps:

> If the new expiration time is not `NaN`, let _expiration time_ be the _(or that?)_ new expiration time.

jdsmith3000: That does seem clearer.  I'll update to use "let <var>expiration time</var> be that expiration time" in both places.

encrypted-media-respec.html
```diff
@@ -1811,7 +1810,7 @@
                       <p class="note">In other words, do not create a session if a non-closed persistent session already exists for this <var>sanitized session ID</var> in any browsing context.</p>
                     </li>
                     <li><p>Load the <var>session data</var>.</p></li>
-                    <li><p>If the <var>session data</var> indicates an expiration time for the session, let <var>expiration time</var> be the expiration time in milliseconds since 01 January 1970 UTC.</p></li>
+                    <li><p>If the <var>session data</var> indicates an expiration time for the session, let <var>expiration time</var> be that expiration time.</p></li>
```
ddorwin: I had assumed that we would explicitly reference the definition for clarity and normative specification.

Options include:
* Appending something like `expressed as <a href="#time">Time</a>`
* `s/that expiration time/that <a href="#time">Time</a>/`

encrypted-media-respec.html
```diff
@@ -579,15 +579,14 @@
 
         <dt id="time">Time</dt>
         <dd>
-          <p>Time represents an instant in time with millisecond accuracy.  The instants in time it can represent are the same that can be represented with 
-            <a class="external" href="https://tc39.github.io/ecma262/#sec-date-objects">ECMAScript <span class="estype">Date</span> objects</a> (ECMA-262, section 20.3 [[!ECMA-262]]) – 
-            namely, every millisecond in the 200,000,000 days centered around midnight of 1 January, 1970 UTC, except for any millisecond that is part of an inserted leap second.  Time does not include these added leap milliseconds.
+          <p>Time MUST be equivalent to that represented in <a class="external" href="https://tc39.github.io/ecma262/#sec-time-values-and-time-range">ECMAScript <span class="estype">Time Values and Time Range</span></a> (ECMA-262, section 20.3.1.1 [[!ECMA-262]]).
```
ddorwin: I worry that the section number will rot, especially since the URL target is a living document. Also, we don't call out sections or even the name of documents elsewhere (except by ``[[foo]]``). Is there a reason you did it this way?

encrypted-media-respec.html
```diff
@@ -579,15 +579,14 @@
 
         <dt id="time">Time</dt>
         <dd>
-          <p>Time represents an instant in time with millisecond accuracy.  The instants in time it can represent are the same that can be represented with 
-            <a class="external" href="https://tc39.github.io/ecma262/#sec-date-objects">ECMAScript <span class="estype">Date</span> objects</a> (ECMA-262, section 20.3 [[!ECMA-262]]) – 
-            namely, every millisecond in the 200,000,000 days centered around midnight of 1 January, 1970 UTC, except for any millisecond that is part of an inserted leap second.  Time does not include these added leap milliseconds.
+          <p>Time MUST be equivalent to that represented in <a class="external" href="https://tc39.github.io/ecma262/#sec-time-values-and-time-range">ECMAScript <span class="estype">Time Values and Time Range</span></a> (ECMA-262, section 20.3.1.1 [[!ECMA-262]]). 
           </p>    
           <p>Time will equal <code>NaN</code> if no such time exists or if the time is indeterminate.  It should never have the value <code>Infinity</code>.
-          </p>
```
ddorwin: We appear to be missing a `</p>` now.

encrypted-media-respec.html
jdsmith3000: That section number was already in the spec.  I just left it.  I think we should just remove it.  The link goes to the specific section anyway.

ddorwin: Ah, yes. I missed that. It has been this way from the beginning: https://github.com/w3c/encrypted-media/commit/41644ebc6606f0fe3a6d689bf0c6e42ebde60321. I think we can remove it.

encrypted-media-respec.html
jdsmith3000: I searched for "millisecond", but somehow missed this one.  I recommend:

>If the <var>session data</var> indicates an expiration time for the session, let <var>expiration time</var> be the expiration time expressed as <a href="#time">Time</a>.

Expiration time shows up 17 times, but this is one of the first uses in the spec, and the only one left with pseudo time quality wording still in it.

ddorwin: This one was fixed (removing millisecond); it just doesn't refer to the definition of Time. I believe your suggestion is consistent with the first option I suggested, and LGTM. The one difference is that you changed "that" back to "the." I'm not sure if that was intentional; "that" seemed to more precisely reference the one we already referenced. This style also simplifies line 2328.

encrypted-media-respec.html
```diff
@@ -2326,7 +2325,7 @@
           <ol>
             <li><p>Let the <var>session</var> be the associated <a>MediaKeySession</a> object.</p></li>
             <li><p>Let <var>expiration time</var> be <code>NaN</code>.</p></li>
-            <li><p>If the new expiration time is not <code>NaN</code>, let <var>expiration time</var> be the new expiration time in milliseconds since 01 January 1970 UTC.</p></li>
+            <li><p>If the new expiration time is not <code>NaN</code>, let <var>expiration time</var> be that expiration time.</p></li>
```
ddorwin: Make the same change here.

encrypted-media-respec.html
ddorwin: We should either use lowercase "time" or update https://w3c.github.io/encrypted-media/encrypted-media-respec.html#dom-mediakeysession-expiration, which uses lowercase "time."

encrypted-media-respec.html
ddorwin: This is perhaps the only place where we really need to reference Time since it is the only place where it is visible. However, it doesn't hurt to do this any place that we get a time.

If we are going to include a reference at line 1813, we also need to do so in:
* the `update()` algorithm
* the Monitor for CDM State Changes algorithm

ddorwin: Also in the the Monitor for CDM State Changes algorithm, please change the _expiration_ variable to _expiration time_ for consistency with all other variables.

ddorwin:
I filed #425 to define "expiration time." I can do that tomorrow if there is agreement.

encrypted-media-respec.html
ddorwin: Actually, #425 might address the reference to Time without needing to add anything to line 1813 or elsewhere (other than perhaps here).

ddorwin:
PR #426 addresses all my comments. If we decide to merge that, there is nothing else to do.

jdsmith3000:
PR #426 is approved for check-in.

PR #426, opened 2017-06-16
Fix #425: Provide a definition for expiration time

### LINKED ISSUE #425: Provide a definition for expiration time
The term "expiration time" is used multiple times in the algorithms (e.g., "If _cdm_ has changed the expiration time of _session_...") yet there is no definition for it.

The definition of the [`expiration`](https://w3c.github.io/encrypted-media/encrypted-media-respec.html#dom-mediakeysession-expiration) attribute says:
>The time after which the key(s) in the session will no longer be usable for decryption, or NaN if no such time exists or if the license explicitly never expires, as determined by the CDM.

While this provides a definition, the relationship to the instances in the algorithm is indirect, especially since those algorithms do not directly modify `expiration`.

We should define "Expiration Time" in the Definitions section and move the first part of the sentence there. We would then have:
>**Expiration Time**
The time after which key(s) will no longer be usable for decryption.

>`expiration`
"The **expiration time** for key(s) in the session, or NaN if no such time exists or if the license explicitly never expires, as determined by the CDM.

>If _cdm_ has changed the **expiration time** of _session_...

ddorwin:
Also address comments on PR #424.

### DIFF
--- a/encrypted-media-respec.html
+++ b/encrypted-media-respec.html
@@ -288,7 +288,7 @@
             Known keys are exposed via the <a def-id="keyStatuses"></a> attribute.
           </p>
 
-          <p>Keys are considered known even after they become unusable, such as due to expiration or if they are removed but a <a def-id="record-of-license-destruction"></a> is available.
+          <p>Keys are considered known even after they become unusable, such as due to <a def-id="expiration-time">expiration</a> or if they are removed but a <a def-id="record-of-license-destruction"></a> is available.
             Keys only become unknown when they are explicitly removed from a session and any license release message is acknowledged.
           </p>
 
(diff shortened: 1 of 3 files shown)


encrypted-media-respec.html
```diff
@@ -2337,7 +2345,7 @@
             <li><p>Let the <var>session</var> be the associated <a>MediaKeySession</a> object.</p></li>
             <li><p>Let <var>expiration time</var> be <code>NaN</code>.</p></li>
             <li><p>If the new expiration time is not <code>NaN</code>, let <var>expiration time</var> be that expiration time.</p></li>
-            <li><p>Set the <var>session</var>'s <a def-id="expiration"></a> attribute to <var>expiration time</var>.</p></li>
+            <li><p>Set the <var>session</var>'s <a def-id="expiration"></a> attribute to <var>expiration time</var> expressed as <a href="#time">time</a>.</p></li>
```
ddorwin: This new text isn't strictly necessary, but it might be nice to reference the time definition again where we expose it to the app. A NOTE might be more appropriate but would probably be distracting.

jdsmith3000:
Please add an updated index.html before you merge.

PR #427, opened 2017-06-23
Fix #406: Unencrypted In-band Support Content

### LINKED ISSUE #406: Note in 8.7.3 Unencrypted In-band Support Content
The Note in section 8,.7.3 is confusing and loosing the intent of the specification. Most of the time, those tracks don't encrypted anyway. How about, replacing the note with the following text and move it after the second sentence:
[[
Note: While in practice accessibility tracks are rarely encrypted, this provision ensures that even in the case that such tracks might be encrypted, the accessibility information will be available to users once the media is decrypted.
]]

This replacement language clarifies how provision 8.7.3 helps ensure access to accessibility info, rather than raising the concern that info won't be available, since testing has demonstrated that the info does remain available.

jdsmith3000:
Wording matches that discussed in the bug.  Creating PR in case there's still time to accept the change.

### DIFF
--- a/encrypted-media-respec.html
+++ b/encrypted-media-respec.html
@@ -3470,7 +3470,8 @@ interface MediaEncryptedEvent : Event {
             Thus, encrypting such tracks would prevent them from being widely available for use with accessibility features in user agent implementations.
           </p>
           <p>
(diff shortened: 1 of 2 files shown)


mwatson2:
LGTM

PR #429, opened 2017-07-28
Replace "user agent" with more accurate terminology

casey:
Given the aim of the Encrypted Media Extensions, it seems prudent to replace "user agent" with more accurate terminology.


casey:
Marked as non substantive for IPR from ash-nazg.

PR #432, opened 2017-10-04
Fix #364: Integrate EME with Feature Policy

### LINKED ISSUE #364: Require explicit enabling of EME in nested contexts
Nested contexts/iframes should only be able to access EME if the embedding app explicitly enables it. The reasons are similar to other features that [will] have such limitations. Specifically, this helps mitigate many security and privacy concerns, especially where the top-level context is not complicit. This includes some of the concerns in #101.

[Feature Policy](https://github.com/wicg/feature-policy/) appears to be the way forward for these purposes.

The default policies would be:
* Enable: `self` for top-level browsing context, and `null` for nested browsing context
* Disable: `null`

The changes to the EME spec itself would likely be similar to those [proposed for Web MIDI](https://github.com/WICG/feature-policy/issues/2). Basically, the promise returned by `requestMediaKeySystemAccess()` would be rejected with `SecurityError` if EME is disabled.

raymeskhoury:
This adds a section defining the feature name and default allowlist, as well as integration with the requestMediaKeySystemAccess algorithm.

***
[Preview](https://s3.amazonaws.com/pr-preview/raymeskhoury/encrypted-media/master.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/encrypted-media/1f8b463...raymeskhoury:e4287db.html)


raymeskhoury:
@clelland we still need to fix the "allowed to use" monkey-patch to make it take a policy-controlled feature. Is that something you could do?

@ddorwin ptal - I'm sure I've violated some style conventions but help would be appreciated in how to do things right - I'm not very experienced with this stuff!

encrypted-media-respec.html
```diff
@@ -657,9 +669,10 @@
 
             <ol class="method-algorithm">
               <!-- TODO: Convert all parameters to use <code>. -->
+              <li><p>Let <var>document</var> be the calling context's <a def-id="document-concept"></a>.</p></li>
+              <li><p>If <var>document</var> is not <a data-cite="!HTML#allowed-to-use">allowed to use</a> the <a data-cite="!FEATURE-POLICY#policy-controlled-feature">policy-controlled feature</a> named <a>encrypted-media</a>, return a promise rejected with a newly created <a def-id="NotSupportedError"></a>.</p></li>
```
raymeskhoury: This could be a SecurityError as recommended in https://github.com/WICG/feature-policy/blob/gh-pages/integration.md. I actually think SecurityError is more appropriate but there may be backward compatibility concerns.

raymeskhoury:
@ddorwin ping :)

encrypted-media-respec.html
ddorwin: I think that's probably fine. NotSupportedError is an expected result that needs to be handled, but nothing is going to work in this case anyway.

@xhwang-chromium, we'll need to update Chromium if this changes (along with the actual algorithm steps).

encrypted-media-respec.html
```diff
@@ -620,6 +620,18 @@
       </dl>
     </section>
 
+    <section>
```
ddorwin: This appears to be a new top-level section, which seems a bit odd. Also, it is specific to a single method, which is defined in the next top-level section. Should it be part of that? How do other APIs handle this?

raymeskhoury: I wasn't sure where to put it some I'm open to suggestions. I don't think there is any standard place right now but @clelland may have thoughts. I'll move it to the next section.

encrypted-media-respec.html
raymeskhoury: Sorry just to clarify: which do you prefer - NotSupportedError or SecurityError?

raymeskhoury:
Thanks - please let me know how to update index.html.

I'm happy for you to do the merge if it's easier.

encrypted-media-respec.html
ddorwin: I am fine with `SecurityError` if that is what other specs are using.

ddorwin:
See https://github.com/w3c/encrypted-media/blob/master/TEAM.md#updating-indexhtml for how to update index.html.

raymeskhoury:
Updated index.html. Note that it changed massively. I'm not sure if I did something wrong or if it is a new version of respec or something. But it changed just as much when I produced a snapshot from top-of-tree. Let me know if I'm missing something!

plehegar:
(to answer David's question)

plehegar:
Feature Policy is still being developed so isn't yet suitable as a normative reference in a W3C Recommendation, so we wouldn't be able to push a new Edited Recommendation with it.
Once it becomes more stable, it's only adding a restriction so, from an IPR perspective, that's fine (cc @wseltzer ). As long as the change gets properly reviewed, we'd be good to go.

encrypted-media-respec.html
foolip: https://github.com/w3c/web-platform-tests/pull/7889 assumes SecurityError.

Malvoz:
Friendly bump, other W3C recommendations such as https://www.w3.org/TR/payment-request/#feature-policy are now integrating this.

joeyparrish:
Any objections to my merging this now (with an update to index.html)?

ddorwin:
It looks like there is an open question of NotSupportedError vs SecurityError in the inline comments.

joeyparrish:
> It looks like there is an open question of NotSupportedError vs SecurityError in the inline comments.

The latest from that discussion was you saying you were fine with SecurityError and foolip saying that WPT expects it.  So I assumed that was settled.

joeyparrish:
JFYI, the "as defined by" part was being rejected by respec as it couldn't resolve the reference to "Feature Policy".  I couldn't figure out the syntax to fix it, but I realized it was redundant, since the sentence already links to the definition in the Feature Policy spec.  So with that removed and index.html regenerated, this merges cleanly.

I'll leave this open for another day to see if anybody has further comment.  If not, I'll merge it.  If I merge it and someone comes along later and says I missed something, we'll fix it.

Thanks!

encrypted-media-respec.html
ddorwin: Updated link: https://github.com/w3c/webappsec-feature-policy/blob/master/integration.md#referencing-the-feature-policy-spec

joeyparrish:
> Updated link: https://github.com/w3c/webappsec-feature-policy/blob/master/integration.md#referencing-the-feature-policy-spec

Thanks!  I've updated my local copy to reflect all of this, including the switch to SecurityError.

joeyparrish:
Since my local diff against master is fairly small and easy to verify against the latest referencing guidelines, I'll go ahead and merge it now.  I'd be happy to follow-up on any changes I missed, or review another PR if someone wants to make additional changes for this.

Thank you @raymeskhoury for your contribution!

PR #444, opened 2018-09-11
Wait for key: run the algorithm when value is false

calvaris:
The second step of the algorithm bails out if the value is true but in
the HTMLMediaElement extensions it is required to run the algorithm
when its value is true, which is useless. Therefore the common sense
requires that it is run when the value is false.


calvaris:
This is wrong, closing

PR #472, opened 2020-07-17
Add note on first/latest decryption time when key never used

### LINKED ISSUE #470: Clarify `first decryption time` when the session is never used
The persistent-usage-record session, the **record of key usage** consists of:

> ...
> **first decryption time** - The first decryption time, defined as the time at which the session was first used to decrypt content, accurate to key usage accuracy and...

It's unclear about the case where the session was never used to decrypt content. In that case, I assume we do not need first decryption time in the record of key usage?

Then in 9.1.5 License Release Format (for Clear Key)

> ...
> For sessions of type "persistent-usage-record" the object shall also contain the following members:
> "firstTime"
> The first decryption time expressed as a number giving the time, in milliseconds since 01 January, 1970 UTC.
> "latestTime"
> The latest decryption time expressed as a number giving the time, in milliseconds since 01 January, 1970 UTC.

If the session was never used to decrypt content. Do we still need "firstTime"? Or shall we have a "firstTime" with value 0?

xhwang-chromium:
Fixes #470

***
<a href="https://pr-preview.s3.amazonaws.com/xhwang-chromium/encrypted-media/pull/472.html" title="Last updated on Jul 17, 2020, 8:12 PM UTC (120a234)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/encrypted-media/472/29db5e9...xhwang-chromium:120a234.html" title="Last updated on Jul 17, 2020, 8:12 PM UTC (120a234)">Diff</a>


joeyparrish:
Merged as a6d383032227fea2717e90d394e62e77efe4b0d4.  Not sure why GitHub didn't mark the PR as merged, though.  I must have flubbed something on the command line.

Thanks for the PR!

PR #479, opened 2020-12-23
Auto-generate respec HTML with GitHub Actions

joeyparrish:
Automatically generate the respec HTML on every commit, using respec2html to automate the process.  Then push the results to the gh-pages branch.

This cuts down on the effort required to make edits to the spec, and avoids issues where we may forget to update index.html or one of the other numerous generated HTML files.

The generated files are now no longer stored in the master branch of the repo, and will only appear in the gh-pages branch.


joeyparrish:
Since this drops the generated HTML from the master branch, we can't merge this until I can talk the repo admins into either:

1. Switching GitHub Pages over to the gh-pages branch
2. Or making me an admin here to I can do it myself

This has all been tested on my fork.  No unexpected changes show up in the generated HTML, and the automation through GitHub is working well.

saschanaz:
Uhm... Didn't respec2html work? 👀

joeyparrish:
I didn't know it existed.  I seem to have wasted a lot of time today.  😁

I saw that the WebDriver spec had a script to do something like this, assumed there was no better tooling available for respect, and rolled up my sleeves to build on that approach.

The good news is that the GitHub Actions workflow won't change much.  I'll just get to drop the Python script and WebDriver.

tidoust:
> Or making me an admin here to I can do it myself

Done!

joeyparrish:
Thanks, @tidoust!

PR #487, opened 2021-06-02
Add reason for session close

### LINKED ISSUE #473: Handling hardware context reset
## Problem

During a playback with hardware-backed media pipeline and CDM, if the device enters sleep or hibernate mode and then resumes, or in some cases when user switches monitors, the hardware context could be completely lost. In this case the playback cannot continue.

On different systems with different media pipeline and CDM implementations, the observable behavior to the JavaScript player could be very different. It could be a `MediaError` on the media element with some platform-specific `MediaError.message`, a zombie `MediaKeySession` on which all subsequent EME calls rejected, a closed `MediaKeySession`, or some combination of them.

This makes it hard for the JavaScript player to react in an interoperable way to provide a good user experience.

## Proposal

- Change EME `MediaKeySession` `closed` attribute to return a `Promise<Reason>` instead of `Promise<void>`, where `Reason` is an enum, e.g. `HARDWARE_CONTEXT_RESET`. This change is backward compatible with the current `closed` attribute that returns `Promise<void>`.
- The media element will fire a `waitingforkey` event and the playback may stall.
- The JavaScript player should handle the session `closed` event(s) to resume playback. For example:
  - Recreate `MediaKeySession` and obtain the license. Once the keys are usable, playback will resume automatically, or
  - Restart the whole playback, including creating a new media element and an `MediaKeys` object.

## Notes

The current EME spec already mentioned that the CDM can close a session at any point:

> The CDM may close a session at any point, such as when the session is no longer needed or when system resources are lost. In that case, the Monitor for CDM Changes algorithm detects the change and runs this algorithm.

However, this is only in a _NOTE_ which is often overlooked by CDM or JavaScript developers. We should explicitly document the logic around spontaneous session close and hardware context reset, e.g., via a dedicated section and/or with an example.

joeyparrish:
This adds a `MediaKeySessionCloseReason` enum, values of which are
used to resolve the `closed` `Promise` on `MediaKeySession`.  This
gives applications additional information about the reasons for a
session closure, which should help applications make better decisions
when a session is closed by the CDM.

Closes #473

***
<a href="https://pr-preview.s3.amazonaws.com/joeyparrish/encrypted-media/pull/487.html" title="Last updated on Jun 24, 2021, 7:55 PM UTC (8617192)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/encrypted-media/487/4c84c08...joeyparrish:8617192.html" title="Last updated on Jun 24, 2021, 7:55 PM UTC (8617192)">Diff</a>

### DIFF
--- a/encrypted-media-respec.html
+++ b/encrypted-media-respec.html
@@ -1389,7 +1389,7 @@ The <dfn>MediaKeysRequirement</dfn> enumeration is defined as follows:
               </div>
               <div class="note">
                 <p>
-                  NOTE: The empty string is distinct from null or not present,
+                  The empty string is distinct from null or not present,
                   and so would be treated as an unrecognized encryption scheme.
                 </p>
               </div>
@@ -1669,7 +1669,7 @@ The <dfn>MediaKeysRequirement</dfn> enumeration is defined as follows:
           <p>The following step is run:</p>
           <ol>
             <li>
-              <p>For each <a>MediaKeySession</a> created by the <var>media keys</var> that is not <a def-id="media-key-session-closed"></a>, <a def-id="queue-a-task"></a> to run the <a def-id="session-closed-algorithm"></a> algorithm on the session.</p>
+              <p>For each <a>MediaKeySession</a> created by the <var>media keys</var> that is not <a def-id="media-key-session-closed"></a>, <a def-id="queue-a-task"></a> to run the <a def-id="session-closed-algorithm"></a> algorithm on the session with reason <a def-id="reason-cdm-unavailable"></a>.</p>
             </li>
           </ol>
         </section>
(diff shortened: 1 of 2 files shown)


encrypted-media-respec.html
```diff
@@ -1732,13 +1732,53 @@ <h2><dfn>MediaKeySession</dfn> Interface</h2>
...
+    "resource-evicted"
+};</pre>
+<p>
+The <dfn>MediaKeySessionClosedReason</dfn> enumeration is defined as follows:
+</p>
+<table class="simple" data-dfn-for="MediaKeySessionClosedReason" data-link-for="MediaKeySessionClosedReason"><tbody><tr><th colspan="2">Enumeration description</th></tr><tr><td><dfn><code id="idl-def-MediaKeySessionClosedReason.unknown">internal-error</code></dfn></td><td>
+            The session was closed because of an unrecoverable error in the CDM.
+            <div class="note">
+              <p>
+                When this occurs, applications SHOULD NOT expect to be able to create new sessions on this <a>MediaKeys</a> instance.
```
jyasskin, line 1749: Drive-by nit: Don't put normative "SHOULD NOT" language inside notes. https://respec.org/docs/#conformance-section claims that Respec flags this, but I don't see an error in your preview.

You could flip the wording to "This MediaKeys instance might not be able to host new sessions." or similar, or elevate the whole thing to normative as "Applications SHOULD NOT attempt to create ...". I wouldn't use "SHOULD NOT expect", as that's making normative requirements on an author's mental state, which is weird. :)

joeyparrish: Thank you for this!  The language of spec is still new for me.

joeyparrish:
@gregwfreedman, please review when you have time.

gregwfreedman (approved):
Looks good!

PR #508, opened 2023-07-24
Update to Node16

gregwfreedman:
ReSpec install is installing puppeteer 20.9 which uses nullish coalescing assignment (??=) which is incompatible with with Node 14. Updating to Node 16 to see if that fixes respec2html.

### DIFF
--- a/.github/workflows/generate-respec-html.yml
+++ b/.github/workflows/generate-respec-html.yml
@@ -12,7 +12,7 @@ jobs:
       - name: Setup Node
         uses: actions/setup-node@v2
         with:
-          node-version: '16'
+          node-version: '14'
       - name: Install Respec
         run: |
           echo '{}' > package.json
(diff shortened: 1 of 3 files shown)


gregwfreedman:
@joeyparrish I am not sure this will work. Is there a way to test the Generate ReSpec HTML Workflow locally?

joeyparrish:
Unfortunately, there is no good way to locally run workflows in a way that accurately represents what will happen on GitHub.  (There are some docker-based things that get close for certain workflows, but I've given up on them because they absolutely fail for more complex workflows.)

The best way to test workflow changes is generally by running the actions in your fork.  So, for example, push this to `main` in your fork (instead of the `node16` branch) and let the "push to main" trigger run there.

But your change looks very safe to me, so in this particular case, let's just merge it and see.

Issue #311, opened 2016-09-07
Early detection of output protection status


mwatson2:
At any given time, Output Protection (HDCP etc.) may be inapplicable (internal displays), not engaged or engaged at various different security levels. The type of video content that may be provided, specifically the resolution, may be dependent on the Output Protection status.

In EME, this may be implemented by providing separate keys for each resolution and a license which specifies the Output Protection requirements for each key. The CDM is responsible for enforcing these license restrictions, including restricting video resolution according to the license conditions.

A problem with the EME API is that the information about Output Protection status, in the form of the status of each key, is not available until after the license is provided.

Unless the CDM supports downscaling, the consequence is one of the following:
- only low resolution video can be downloaded before the license has been provided. The presentation begins at low quality.
- downloading of video data cannot begin before the license has been provided. The presentation takes longer to start because of this.
- high resolution data downloaded before the license has been provided may need to be discarded once the license has been provided and key status determined. The presentation takes longer to start in these cases and data is downloaded unnecessarily.

Since downscaling is not widely supported, this feature requests asks for a mechanism to discover output protection status before the license is provided, without a round trip to the server.

mwatson2:
Without pre-empting discussion of this feature request, I would like to briefly document a few possible solutions (without comment on their merits):

- A new API could be provided which returns the Output Protection status, for example one of { <internal display>, <not engaged>, <hdcp 1.4 engaged>, <hdcp 2.2 engaged>, ... }.
- Output Protection could be considered a form of robustness and values for the `robustness` field in `requestMediaKeySystemAccess` could be defined to represent different Output Protection reuiqrements.
- Output Protection requirements could be specified early in `initData` allowing key status to be exposed at the same time the key request message is generated.

joeyparrish:
It seems to me that discovering output protection status early only helps if you know what the policy is for each resolution.  If you don't know the output protection policy per resolution, you must wait for the license anyway.  Perhaps this is easy for some applications that are in control of the entire stack, but for generic apps or middleware, this is not a likely scenario.

I strongly object to putting output protection requirements in init data.  Output protection requirements are part of license policies, and policies can be changed on the license server at any time.  Putting policy information of any kind into init data means the data the client has may be out of sync.  Once players are bitten by this a time or two, they will quickly start ignoring any such information.

mwatson2:
[...]

PR #515, opened 2023-10-16
Add getStatusForPolicy

gregwfreedman:
Add the method `getStatusForPolicy` as described in [Encrypted Media: HDCP Policy Check](https://wicg.github.io/hdcp-detection/) to address #311.

***
<a href="https://pr-preview.s3.amazonaws.com/gregwfreedman/encrypted-media/pull/515.html" title="Last updated on Nov 10, 2023, 4:40 PM UTC (d3c0565)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/encrypted-media/515/4089d85...gregwfreedman:d3c0565.html" title="Last updated on Nov 10, 2023, 4:40 PM UTC (d3c0565)">Diff</a>

### DIFF
--- a/encrypted-media-respec.html
+++ b/encrypted-media-respec.html
@@ -1659,7 +1659,7 @@ The <dfn>MediaKeysRequirement</dfn> enumeration is defined as follows:
                   </tr>
                 </tbody>
               </table>
-              <div><em>Return type: </em><code>MediaKeyStatus</code></div>
+              <div><em>Return type: </em><code>Promise&lt;MediaKeyStatus&gt;</code></div>
 
               <div>
                 <pre class="idl">
--- a/encrypted-media-respec.html
+++ b/encrypted-media-respec.html
@@ -1650,7 +1650,7 @@ The <dfn>MediaKeysRequirement</dfn> enumeration is defined as follows:
                       <code>MediaKeysPolicy</code>
                     </td>
                     <td class="prmNullFalse">
-                      <span role="img" aria-label="True">✔</span>
+                      <span role="img" aria-label="False">✘</span>
                     </td>
                     <td class="prmOptFalse">
                       <span role="img" aria-label="True">✔</span>
@@ -4845,6 +4845,23 @@ interface MediaEncryptedEvent : Event {
 
 &lt;video id='v' src='foo.webm' autoplay&gt;&lt;/video&gt;
 </pre>
+      </section>
+
+      <section id="example-getStatusForPolicy">
+        <h3>Pre-fetch Media With HDCP Policy</h3>
+        <p class="exampledescription">If some media will have an HDCP policy in the license, an application can check for that version before pre-fetching.</p>
+
+        <pre class="example highlight">
+          const status = await video.mediaKeys.getStatusForPolicy({
+            minHdcpVersion: '1.4'
+          });
+          
+          if (status === 'usable') {
+            // Pre-fetch HD content.
+          } else {  // 'output-restricted'
+            // Pre-fetch SD content.
+          }
+        </pre>
       </section>
     </section>
 
--- a/encrypted-media-respec.html
+++ b/encrypted-media-respec.html
@@ -1564,13 +1564,20 @@ The <dfn>MediaKeysRequirement</dfn> enumeration is defined as follows:
         </td></tr>
       </tbody></table></div>
 
-      <div><pre class="idl">[Exposed=Window, SecureContext] interface MediaKeys {
-    MediaKeySession  createSession (optional MediaKeySessionType sessionType = "temporary");
-    Promise&lt;boolean&gt; setServerCertificate (BufferSource serverCertificate);
-};</pre><section><h2>Methods</h2><dl class="methods" data-dfn-for="MediaKeys" data-link-for="MediaKeys"><dt><dfn><code>createSession</code></dfn></dt><dd>
+      <div>
+        <pre class="idl">
+          [Exposed=Window, SecureContext] interface MediaKeys {
+              MediaKeySession  createSession (optional MediaKeySessionType sessionType = "temporary");
+              Promise&lt;MediaKeyStatus&gt; getStatusForPolicy (optional MediaKeysPolicy policy = {});
+              Promise&lt;boolean&gt; setServerCertificate (BufferSource serverCertificate);
+          };
+        </pre>
+
+        <section id="">
+          <h2>Methods</h2>
+          <dl class="methods" data-dfn-for="MediaKeys" data-link-for="MediaKeys">
+            <dt><dfn><code>createSession</code></dfn></dt><dd>
           <p>Returns a new <a>MediaKeySession</a> object.</p>
-
-
 
           <ol class="method-algorithm">
             <li><p>If this object's <var>supported session types</var> value does not contain <var>sessionType</var>, <a def-id="throw"></a> [[WEBIDL]] a <a def-id="NotSupportedError"></a>.</p>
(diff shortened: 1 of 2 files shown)


joeyparrish (approved):
LGTM.  @xhwang-chromium, any feedback?

chrisn:
LGTM too. 

The WICG draft also contains some security and privacy considerations and a code example that we might want to consider including. There's also a question there about whether we should introduce a registry for `HDCPVersion` to allow updates to the enum without changing the EME spec.

We can follow up on these as separate issues though.

encrypted-media-respec.html
```diff
@@ -1604,15 +1611,95 @@ <h2><dfn>MediaKeys</dfn> Interface</h2>
...
+                      <span role="img" aria-label="True">✔</span>
+                    </td>
+                    <td class="prmOptFalse">
+                      <span role="img" aria-label="True">✔</span>
+                    </td>
+                    <td class="prmDesc">A set of policy requirements</td>
+                  </tr>
+                </tbody>
+              </table>
+              <div><em>Return type: </em><code>MediaKeyStatus</code></div>
```
xhwang-chromium, line 1662: Should this be `Promise<MediaKeyStatus>`?

encrypted-media-respec.html
```diff
@@ -1604,15 +1611,95 @@ <h2><dfn>MediaKeys</dfn> Interface</h2>
...
+                  <tr>
+                    <td class="prmName">policy</td>
+                    <td class="prmType">
+                      <code>MediaKeysPolicy</code>
+                    </td>
+                    <td class="prmNullFalse">
+                      <span role="img" aria-label="True">✔</span>
+                    </td>
+                    <td class="prmOptFalse">
+                      <span role="img" aria-label="True">✔</span>
```
xhwang-chromium, line 1656: Why do we allow the `policy` to be `Nullable` and `Optional`? It seems simpler not to allow them. If we do allow, the spec should handle these cases.

encrypted-media-respec.html
```diff
@@ -1604,15 +1611,95 @@ <h2><dfn>MediaKeys</dfn> Interface</h2>
...
+                  enum HDCPVersion {
+                      "1.0",
+                      "1.1",
+                      "1.2",
+                      "1.3",
+                      "1.4",
+                      "2.0",
+                      "2.1",
+                      "2.2",
+                      "2.3",
```
xhwang-chromium, line 1685: Just a note here. If the industry introduces newer HDCP versions, e.g. "2.4", we'll have to update the EME spec to support it. Historically we've been using Registries to deal with similar cases; I am not sure whether we want to do it this time with the additional complexity.

encrypted-media-respec.html
```diff
@@ -1604,15 +1611,95 @@ <h2><dfn>MediaKeys</dfn> Interface</h2>
...
               The type of session to create. The session type affects the behavior of the returned object.
-            </td></tr></tbody></table><div><em>Return type: </em><code>MediaKeySession</code></div></dd><dt><dfn><code>setServerCertificate</code></dfn></dt><dd>
+            </td></tr></tbody></table><div><em>Return type: </em><code>MediaKeySession</code></div></dd>
+            <dt><dfn><code>getStatusForPolicy</code></dfn></dt>
+            <dd>
+              <p id="status-for-policy">Returns the <a>MediaKeyStatus</a> for a given <a>MediaKeysPolicy</a>.</p>
+
+              <ol class="method-algorithm">
+                <li>If <var>policy</var> has no <a def-id="present-dictionary-member"></a> <a def-id="dictionary-member">members</a>, return a promise rejected with a newly created <a def-id="TypeError"></a>.</li>
```
xhwang-chromium, line 1620: I don't know how the links work in the spec. But there seems to be some issue with the link on "present" and "members", based on the preview page.

The first one points to https://webidl.spec.whatwg.org/#map-exists, which seems to be a broken anchor in the page.

The second one points to https://webidl.spec.whatwg.org/##dfn-dictionary-member, which should be https://webidl.spec.whatwg.org/#dfn-dictionary-member (## -> #).

xhwang-chromium (commented):
Overall it looks pretty good. I left some minor comments for discussion.

encrypted-media-respec.html
gregwfreedman: @tidoust @chrisn Should I do a registry for this? It seems like [https://github.com/w3c/encrypted-media](https://github.com/w3c/encrypted-media) would have to updated in either case.

encrypted-media-respec.html
gregwfreedman: IIUC, the original idea for this method was to make it extensible, by adding other policies to MediaKeysPolicy. If this is where the API is going then it doesn't make sense to make `minHdcpVersion` required. And dictionary argument must be optional if it has no required fields. I don't think it makes sense to make `minHdcpVersion` required now, if it might not be in the future (however unlikely).

xhwang-chromium: Fully agree with what you said! However, the question was about `policy`, not `minHdcpVersion`. The current draft says the `policy` is `Nullable` and `Optional`, and I was asking about that.

encrypted-media-respec.html
chrisn: It does add a bit of extra complexity, but we have examples in WebCodecs we can borrow from that follow the latest Registry track Process requirements, so shouldn't be too difficult. It would mean we can add new versions without updating the spec itself which could be useful once we've completed v2.

encrypted-media-respec.html
gregwfreedman: I actually responded to that. When a dictionary (i.e. the policy) has no required fields then it must be optional/nullable. Respec gives a WebIDL validation error when policy is not optional/nullable.
> And dictionary argument must be optional if it has no required fields.

If you look at [Encrypted Media: HDCP Policy Check](https://wicg.github.io/hdcp-detection/), you will see this is defined the same way. I'm pretty sure that's why @joeyparrish wrote the explainer that way.

xhwang-chromium: Thanks for the explanation! I leaned something today. It feels odd to allow calling this method with null, but I guess you could always call it with an empty dictionary, which is the same thing.

@tidoust to double check on this.

Since the parameter is nullable, in the algorithm, should we handle it?

tidoust: The Web IDL definition looks good to me.

For `optional`, that is indeed required (with a default value).

I don't understand what makes the parameter nullable, though? For it to be nullable, [according to Web IDL](https://webidl.spec.whatwg.org/#idl-nullable-type), it would have to be defined with a `?`, as in:

```
Promise<MediaKeyStatus> getStatusForPolicy (
  optional MediaKeysPolicy? policy = {}
);
```

... and Web IDL also says that dictionary types cannot be nullable when used as the type of an operation argument, so that does not seem allowed in any case.

gregwfreedman: I changed `policy` to NOT `Nullable`.

encrypted-media-respec.html
tidoust: I missed this comment a couple of weeks ago. Yes, a registry would seem appropriate, also as a way to reference what "x.y" means in practice with a link to the relevant specification.

If you follow the registry path, I note that you'll need to drop the   `enum` and rather use a `DOMString`, as [done in WebCodecs](https://www.w3.org/TR/webcodecs/#dom-audiodecoderconfig-codec).

gregwfreedman: Will do.

PR #516, opened 2023-11-14
Add draft HDCP version registry

chrisn:
Following from #515, this adds a draft registry for HDCP versions.

Still to do:

- [x] Add references for HDCP 2.0 and upwards
- [x] Add spec links
- [ ] Update EME spec to reference the registry

### DIFF
--- a/hdcp-version-registry-respec.html
+++ b/hdcp-version-registry-respec.html
@@ -19,30 +19,101 @@
       group: "media",
       localBiblio: {
         "HDCP-1.0": {
-          "title": "High-bandwidth Digital Content Protection System, Revision 1.0",
-          "publisher": "Digital Content Protection LLC",
-          "date": "17 Feburary 2000"
+          title: "High-bandwidth Digital Content Protection System, Revision 1.0",
+          publisher: "Digital Content Protection LLC",
+          date: "17 Feburary 2000",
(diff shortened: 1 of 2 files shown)


hdcp-version-registry-respec.html
```diff
@@ -0,0 +1,175 @@
...
+
+      // the specification's short name, as in https://www.w3.org/TR/short-name/
+      shortName: "eme-hdcp-version-registry",
+
+      // if there a publicly available Editor's Draft, this is the link
+      edDraftURI: "https://w3c.github.io/encrypted-media/hdcp-version-registry.html",
+
+      // if this is a LCWD, uncomment and set the end of its review period
+      // lcEnd: "2009-08-05",
+
```
tidoust, line 20: We don't really do Last Call Working Drafts anymore, so no need to keep these comments around, `lcEnd` should not be of any use.

```suggestion
```

chrisn: I have removed this. Thanks!

joeyparrish (approved):

gregwfreedman (approved):

chrisn:
Thank you all for reviewing. I suggest making a new PR to update the EME spec to reference the registry, so will go ahead and merge this.

PR #524, opened 2024-03-14
Update registries to use W3C Registry Track

chrisn:
This PR updates the EME initdata and Stream Format Registries to use the [W3C Registry Track](https://www.w3.org/2023/Process-20231103/#registries).

* ReSpec complains that there should be a compliance section. The [WebCodecs Codec Registry](https://www.w3.org/TR/webcodecs-codec-registry/) doesn't have one, so I'm not sure what to do about the warning.
* "Initialization Data Type" is referenced by the initdata registry, so I have exported the definition so ReSpec can find it.
* The Stream Format Registry requirements described needing an Initialization Data Type, whereas it actually needs a MIME type/subtype pair. I've copied wording from the [MSE Byte Stream Format Registry](https://www.w3.org/TR/mse-byte-stream-format-registry/)
* For completeness, and as suggested (but not required) by the Registry Track process, I added that existing entries cannot be deleted or deprecated.
* Use the latest ReSpec script, and removed the custom JavaScript, which is no longer needed.
* Use the DRY (draft registry) document type.
* Use GitHub issues to add new entries, rather than a mailing list.
* RFC2119 words are sometimes lower case, sometimes upper case, should we make these consistent?

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/encrypted-media/pull/524.html" title="Last updated on Mar 18, 2024, 8:30 PM UTC (4b23f67)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/encrypted-media/524/0bc7d34...4b23f67.html" title="Last updated on Mar 18, 2024, 8:30 PM UTC (4b23f67)">Diff</a>

### DIFF
--- a/encrypted-media-respec.html
+++ b/encrypted-media-respec.html
@@ -291,7 +291,7 @@
 
         <dt id="decryption-key-id">Key ID</dt>
         <dd>
-          <p>A <a href="#decryption-key">key</a> is associated with a key ID that is a sequence of octets and which uniquely identifies the key.
+          <p>A <a href="#decryption-key">key</a> is associated with a <dfn class="export" data-lt="Decryption key ID">key ID</dfn> that is a sequence of octets and which uniquely identifies the key.
           The container specifies the ID of the key that can decrypt a block or set of blocks within the <a def-id="media-data"></a>.
           <a def-id="initialization-data"></a> MAY contain key ID(s) to identify the keys that are needed to decrypt the media data.
           However, there is no requirement that Initialization Data contain any or all key IDs used in the <a def-id="media-data"></a> or <a def-id="media-resource"></a>.
@@ -3029,7 +3029,7 @@ interface MediaEncryptedEvent : Event {
         <section id="initdata-encountered">
           <h4>Initialization Data Encountered</h4>
           <p>
-            The Initialization Data Encountered algorithm queues an <a def-id="encrypted"></a> event for <a def-id="initialization-data"></a> encounterd in the <a def-id="media-data"></a>.
+            The <dfn class="export">Initialization Data Encountered</dfn> algorithm queues an <a def-id="encrypted"></a> event for <a def-id="initialization-data"></a> encounterd in the <a def-id="media-data"></a>.
             Requests to run this algorithm include a target <a def-id="htmlmediaelement"></a> object.
           </p>
           <p>The following steps are run:</p>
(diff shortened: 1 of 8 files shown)


tidoust:
> ReSpec complains that there should be a compliance section. The [WebCodecs Codec Registry](https://www.w3.org/TR/webcodecs-codec-registry/) doesn't have one, so I'm not sure what to do about the warning.

What bothers ReSpec is that the spec registry entry requirements use RFC 2119 terms (with upper case syntax), which indeed suggests that the spec contains normative statements. The section should be updated to use lower case variants (or other formulations).

> RFC2119 words are sometimes lower case, sometimes upper case, should we make these consistent?

See above, they SHOULD be removed ;)
For ReSpec, making them lower case is enough. For readers, it could be worth reformulating, but that does not strike me as essential.

> Use the DRY (draft registry) document type.

The DRY document type is a /TR status. There is no "Editor's Draft of a (Draft|Candidate) Registry" in ReSpec. The WebCodecs Registry sticks to ED there. The right status gets set during publication.

chrisn:
I'll update as you suggest. Thanks!

tidoust (approved):
Looks good to me!

chrisn:
I've just realised that the registry entries will also need updating to the latest ReSpec, so I'll add those changes to this PR.

encrypted-media-respec.html
```diff
@@ -291,7 +291,7 @@ <h2>Definitions</h2>
 
         <dt id="decryption-key-id">Key ID</dt>
         <dd>
-          <p>A <a href="#decryption-key">key</a> is associated with a key ID that is a sequence of octets and which uniquely identifies the key.
+          <p>A <a href="#decryption-key">key</a> is associated with a <dfn class="export" data-lt="Decryption key ID">key ID</dfn> that is a sequence of octets and which uniquely identifies the key.
```
tidoust, line 294: Why inline the `<dfn>` when it could go directly to within the `<dt>`?

```html
<dt id="decryption-key-id"><dfn class="export" data-lt="Decryption key ID">Key ID</dfn></dt>
```

(Ideally, the `id` on the `<dt>` would go too, but we'll need to get rid of the `<a def-id="key-id"></a>` constructs first)

encrypted-media-respec.html
```diff
@@ -3029,7 +3029,7 @@ <h4>Media Data May Contain Encrypted Blocks</h4>
         <section id="initdata-encountered">
           <h4>Initialization Data Encountered</h4>
           <p>
-            The Initialization Data Encountered algorithm queues an <a def-id="encrypted"></a> event for <a def-id="initialization-data"></a> encounterd in the <a def-id="media-data"></a>.
+            The <dfn class="export">Initialization Data Encountered</dfn> algorithm queues an <a def-id="encrypted"></a> event for <a def-id="initialization-data"></a> encounterd in the <a def-id="media-data"></a>.
```
tidoust, line 3032: Same comment here (and below for "Encrypted Block Encountered"): it seems better to me to have the dfn on the `<h4>` directly:

```html
<h4><dfn class="export">Initialization Data Encountered</dfn></h4>
```

encrypted-media-respec.html
chrisn: Thanks, makes sense. Updated in https://github.com/w3c/encrypted-media/pull/524/commits/4b23f67c6960cb3beee7065987e0e04c81d41228.

encrypted-media-respec.html
chrisn: Done

joeyparrish (approved):

PR #527, opened 2024-03-20
Fix exports

chrisn:
This is a follow up to #524. The attributes are needed because the spec uses an older version of ReSpec (which we can address later). This also adds an export for "Initialization Data", for the MP4 stream registry spec.

### DIFF
--- a/encrypted-media-respec.html
+++ b/encrypted-media-respec.html
@@ -317,7 +317,7 @@
           <p>A license is key system-specific state information that includes one or more <a href="#decryption-key">key(s)</a> - each associated with a <a def-id="key-id"></a> - and potentially other information about key usage.</p>
         </dd>
 
-        <dt id="initialization-data">Initialization Data</dt>
+        <dt id="initialization-data"><dfn class="export" data-export="">Initialization Data</dfn></dt>
         <dd>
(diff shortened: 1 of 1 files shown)


encrypted-media-respec.html
```diff
@@ -332,7 +332,7 @@ <h2>Definitions</h2>
 
           <p>
           The format of the initialization data depends upon the type of container, and containers MAY support more than one format
-          of initialization data. The <dfn class="export" id="initialization-data-type">Initialization Data Type</dfn> is a string that indicates the
```
tidoust, line 335: Can this ID be dropped? I suspect ReSpec will create a different ID (`dfn-initialization-data-type`) and the spec contains references to the term that use the [linking system in `encrypted-media.js`](https://github.com/w3c/encrypted-media/blob/2e4580cabc4bdd150547db7c362a6652d0689a71/encrypted-media.js#L172), which expects the ID to exist.

chrisn: Ah, let me restore this for now, then.

chrisn: Done

tidoust (approved):

PR #535, opened 2024-05-13
Update minHdcpVersion to reference registry

gregwfreedman:
This updates `minHdcpVersion` to point to the registry. [Specref](https://www.specref.org/) will need to be updated as well.

***
<a href="https://pr-preview.s3.amazonaws.com/gregwfreedman/encrypted-media/pull/535.html" title="Last updated on May 13, 2024, 9:53 PM UTC (b1303c0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/encrypted-media/535/c01e7e5...gregwfreedman:b1303c0.html" title="Last updated on May 13, 2024, 9:53 PM UTC (b1303c0)">Diff</a>

### DIFF
--- a/encrypted-media-respec.html
+++ b/encrypted-media-respec.html
@@ -2979,7 +2979,7 @@
             <div>
               <pre class="idl">
                   dictionary MediaKeysPolicy {
-                      HDCPVersion minHdcpVersion;
+                      DOMString minHdcpVersion;
                   };
                 </pre>
               <p>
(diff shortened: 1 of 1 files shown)


joeyparrish (approved):

chrisn (approved):

encrypted-media-respec.html
```diff
@@ -2990,23 +2990,11 @@ <h2>
...
-                  };
-                </pre>
               <p>
-                The HDCP Policy is represented by {{MediaKeysPolicy/minHdcpVersion}}. When set, the
-                policy requirement will be fulfilled if the system supports
-                {{MediaKeysPolicy/minHdcpVersion}} on the current display.
+                The HDCP Policy is represented by {{MediaKeysPolicy/minHdcpVersion}}. If the system
+                can enable the HDCP version specified or higher, then the policy will result in a
+                {{MediaKeyStatus}} of {{MediaKeyStatus/"usable"}}. Implementations MUST support
+                {{MediaKeysPolicy/minHdcpVersion}} values defined in [[EME-HDCP-VERSION-REGISTRY]].
```
tidoust, line 2997: Apologies for the late review.

I'm not sure what that requirement means. A registry evolves over time. If "2.4" gets added to the registry, it seems fine for an implementation not to support the *value* initially, the algorithm will just throw a `NotSupportedError` error in that case.

Also, registries are informative in essence, so a normative reference to them is strange. I would replace with a non-normative "The [[?EME-HDCP-VERSION-REGISTRY]] provides the mapping from {{MediaKeysPolicy/minHdcpVersion}} values to HDCP specifications".

gregwfreedman: Not late at all, happy to get feedback anytime.

I'm happy to make this change, but I was just borrowing the language from elsewhere in the spec, "Implementations MUST support the common formats defined in [[EME-INITDATA-REGISTRY]]..."

chrisn: I just found the relevant part of the Process is https://www.w3.org/2023/Process-20231103/#reg-ref-specifications

gregwfreedman: https://github.com/w3c/encrypted-media/pull/540

tidoust: > I'm happy to make this change, but I was just borrowing the language from elsewhere in the spec, "Implementations MUST support the common formats defined in [[EME-INITDATA-REGISTRY]]..."

Yes, I would say that this needs fixing as well ;) The formats that must be supported need to appear explicitly in the spec itself, as illustrated in the example @chrisn referenced from the Process.

chrisn: OK, so the key question is: are we comfortable with these being non-normative? If not, then adding a registry isn't the right direction. Sorry, my fault, should have gone through this earlier.

PR #538, opened 2024-05-14
Resolve promises before enqueuing events

### LINKED ISSUE #461: Clarify explicitly when promises are to be resolved
Currently the spec notes "Since promise handlers are queued as microtasks, these will be executed ahead of any events queued by the preceding steps" in several places. For example, in [generateRequest](https://w3c.github.io/encrypted-media/#methods-2):

> 10. Queue a task to run the following steps:
> > 1. If any of the preceding steps failed, reject promise with a new DOMException whose name is the appropriate error name.
> > 2. Set the sessionId attribute to session id.
> > 3. Set this object's callable value to true.
> > 4. Run the Queue a "message" Event algorithm on the session, providing message type and message.
> > 5. Resolve promise.

> NOTE: Since promise handlers are queued as microtasks, these will be executed ahead of any events queued by the preceding steps.

To make the spec clearer, it would be better to indicate exactly when the promise is resolved, compared to other steps. The example above would switch 4 and 5 to become:
> 4. Resolve promise.
> 5. Run the Queue a "message" Event algorithm on the session, providing message type and message.

And then the note is not needed, as it is now clear that the promise should be resolved before the "message" event is generated (and does not depend on promise handlers being implemented as microtasks). This also appears to be the suggestion in #19 ([comment](https://github.com/w3c/encrypted-media/issues/19#issuecomment-188497121)) although it was switched back when addressing #14.

Resolving the promise before firing an event is used in other specs (e.g. [Presentation API](https://w3c.github.io/presentation-api/#starting-a-presentation-connection)).

There is a potential problem with this in close(). Currently the attribute closed promise is resolved before the close() method promise is resolved. This will break the WPT tests for [playback-temporary-events](https://github.com/web-platform-tests/wpt/blob/master/encrypted-media/scripts/playback-temporary-events.js#L86) as it checks that the attribute promise is resolved before the close() method promise is resolved.

joeyparrish:
Closes #461

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/encrypted-media/pull/538.html" title="Last updated on Jun 7, 2024, 5:55 PM UTC (8abca03)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/encrypted-media/538/c01e7e5...8abca03.html" title="Last updated on Jun 7, 2024, 5:55 PM UTC (8abca03)">Diff</a>

### DIFF
--- a/encrypted-media-respec.html
+++ b/encrypted-media-respec.html
@@ -3026,6 +3026,11 @@
                 <p>
                   Let <var>promise</var> be a new promise.
                 </p>
+              </li>
+              <li>
+                <p>
+                  [=Queue a task=] to run the following steps:
+                </p>
                 <ol>
                   <li>
                     <p>
--- a/encrypted-media-respec.html
+++ b/encrypted-media-respec.html
@@ -4389,8 +4389,8 @@
             <ol class="method-algorithm">
               <li>
                 <p>
-                  If this object's <var>closing or closed</var> value is true, return a resolved
-                  promise.
+                  If this object's <var>closing or closed</var> value is true, return a promise
+                  resolved with <code>undefined</code>.
                 </p>
               </li>
               <li>
--- a/encrypted-media-respec.html
+++ b/encrypted-media-respec.html
@@ -5703,7 +5703,7 @@
               <li>
                 <p>
                   If <var>mediaKeys</var> and the {{HTMLMediaElement/mediaKeys}} attribute are the
-                  same object, return a resolved promise.
+                  same object, return a promise resolved with <code>undefined</code>.
                 </p>
               </li>
               <li>
--- a/encrypted-media-respec.html
+++ b/encrypted-media-respec.html
@@ -3779,17 +3779,13 @@
                       </li>
                       <li>
                         <p>
-                          Run the [=Queue a "message" Event=] algorithm on the <var>session</var>,
-                          providing <var>message type</var> and <var>message</var>.
+                          Resolve <var>promise</var>.
                         </p>
                       </li>
                       <li>
                         <p>
-                          Resolve <var>promise</var>.
-                        </p>
-                        <p class="note">
-                          Since promise handlers are queued as microtasks, these will be executed
-                          ahead of any events queued by the preceding steps.
+                          Run the [=Queue a "message" Event=] algorithm on the <var>session</var>,
+                          providing <var>message type</var> and <var>message</var>.
                         </p>
                       </li>
                     </ol>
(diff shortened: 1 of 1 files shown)


joeyparrish:
@jrummell-chromium JFYI

chrisn (approved):
@jernoble Does this look OK to you, any potential interop concerns?

marcoscaceres (changes requested):
I don’t think this is correct. Promises need to be resolved in task queues also. Also, promises should resolve with *something* even if that’s an explicit `undefined`.

joeyparrish:
> Promises need to be resolved in task queues also.

I don't understand what you mean by this.  Can you suggest an edit to show me what language you would agree with?

marcoscaceres:
Sorry for the delay: 

For example, from the Fetch spec:
1. [Queue a fetch task](https://fetch.spec.whatwg.org/#queue-a-fetch-task) to run the following steps, with fetchParams’s [task destination](https://fetch.spec.whatwg.org/#fetch-params-task-destination).
1.1 [Pull from bytes](https://streams.spec.whatwg.org/#readablestream-pull-from-bytes) buffer into stream.
1.1 If stream is [errored](https://streams.spec.whatwg.org/#readablestream-errored), then [terminate](https://fetch.spec.whatwg.org/#fetch-controller-terminate) fetchParams’s [controller](https://fetch.spec.whatwg.org/#fetch-params-controller).
1.1 [Resolve](https://webidl.spec.whatwg.org/#resolve) promise with undefined.


See also https://www.w3.org/TR/web-share/ ... how the global task queue is used to resolve promises.

encrypted-media-respec.html
```diff
@@ -3779,17 +3779,13 @@ <h2>
                       </li>
                       <li>
                         <p>
-                          Run the [=Queue a "message" Event=] algorithm on the <var>session</var>,
-                          providing <var>message type</var> and <var>message</var>.
+                          Resolve <var>promise</var>.
```
marcoscaceres, line 3782: Like.. this ... (please don't merge this, just an example):

```suggestion
                          , [queue a global task](https://html.spec.whatwg.org/multipage/webappapis.html#queue-a-global-task) on the [??? task source](https://html.spec.whatwg.org/multipage/webappapis.html#user-interaction-task-source) using global to resolve <var>promise</var> with `undefined`.
```

marcoscaceres: Note that promises must resolved with something... even is it's `undefined`

joeyparrish: Okay, now that I know what to look for, I see lots of examples of "resolve promise with undefined" in other Media WG specs.  Thank you for pointing that out!

I don't see anything like "queue a global task on ... to resolve promise" in any of our other specs.  I see examples of "resolve" in the steps of a queued task, or in the steps of something run "in parallel".  In fact, this bit you're commenting on is already inside of "[=Queue a task=] to run the following steps".  So it seems like this part of your suggestion is maybe already covered.  Please let me know if I misunderstood.

I'll update all the resolutions to include a value, even if it's `undefined`.

joeyparrish:
I audited every instance of "resolve" or "resolved" in the spec.  All have values, and all are now inside either "Run the following steps in parallel" or "queue a task" or "Use the cdm to execute".  (I had to fix only one instance of a missing queue, in the recently-introduced method `getStatusForPolicy()`.)

joeyparrish:
Also worth a note, since we're being thorough: we resolve a promise inside the session closed algorithm, which is always run inside a task or in a set of parallel steps.

joeyparrish:
I'm ready to merge this if there is no further feedback.

I'll be on leave for five weeks starting June 22, so if a positive consensus is reached after that date, please don't wait for me to hit the "merge" button.

encrypted-media-respec.html
marcoscaceres: Yeah, all our specs are wrong 😭 we need to fix them.

marcoscaceres: We can fix things incrementally though. 

They way to understand this is that something in parallel happens off main thread, then with the result of the asynchronous operation we queue a task (on the main thread, using an explicit task queue) to resolve that promise with some value.

gregwfreedman (approved):

PR #553, opened 2024-07-22
Enable auto-publication

tidoust:
This automates the publication of all 9 EME documents to the `gh-pages` branch and to /TR, using the w3c/spec-prod action.

The `W3CTRMANIFEST` files are no longer needed as a result and nothing needs to be Git-ignored anymore either.

Also note the change of shortname for Encrypted Media Extensions to `encrypted-media-2`.

I added the Echidna tokens as repository secrets.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/encrypted-media/pull/553.html" title="Last updated on Jul 23, 2024, 2:11 PM UTC (3af0bfb)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/encrypted-media/553/576c0e1...3af0bfb.html" title="Last updated on Jul 23, 2024, 2:11 PM UTC (3af0bfb)">Diff</a>


chrisn:
Thank you, I have just one question. I notice that /TR/encrypted-media now also shows the FPWD, will both /TR/encrypted-media and /TR/encrypted-media-2 point to the same document as we make more updates? Or should /TR/encrypted-media still point to the existing Rec?

tidoust:
> Thank you, I have just one question. I notice that /TR/encrypted-media now also shows the FPWD, will both /TR/encrypted-media and /TR/encrypted-media-2 point to the same document as we make more updates?

Yes. I requested the same setting as for MSE: the series shortname `encrypted-media` returns the latest level of the specification, meaning `encrypted-media-2`.

> Or should /TR/encrypted-media still point to the existing Rec?

Same as for MSE as well: the previous Recommendation got a (new) `encrypted-media-1` shortname and is available at https://www.w3.org/TR/encrypted-media-1/.

chrisn:
Perfect, that all makes sense.

marcoscaceres:
Will these all run when we change one file? I wonder if we should bind each job to a particular path?

tidoust:
> Will these all run when we change one file?

Yes.

> I wonder if we should bind each job to a particular path?

That would be useful for WebCodecs too (16 jobs run whenever a file is changed).

It's not trivial because the [`paths` filter](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore) runs at the workflow level, whereas we would need it at the job level.

We could create one workflow per document. That works. That's not DRY-friendly. But then that's simple. Thinking about it, it should be trivial to script the creation of the workflow document to ease maintenance and address the "not DRY-friendly" concern.

Or we need to check what changed ourselves within a step. The [Paths Change Filter](https://github.com/marketplace/actions/paths-changes-filter) action seems to provide what we need. To avoid starting a new job for each and every document, we would need to drop the "matrix" approach, and rather create one step per document. We'd lose a bit of DRY-ness as well.

I'll look into it. Let me know if you see other approaches!

tidoust:
I created dedicated workflows with `paths` filters, generated from a companion script and a workflow template that I put under the `.github` folder as well. How does that look, @marcoscaceres?

joeyparrish (approved):

tidoust:
Merging. We can always iterate and improve the workflow afterwards :)

PR #609, opened 2026-06-09
Replace "use the cdm"

chrisn:
Various algorithms in the spec say "use the cdm", but in these cases "cdm" is actally a variable that refers to an instance of the CDM. Hence "use cdm" is more appropriate.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/encrypted-media/pull/609.html" title="Last updated on Jun 9, 2026, 7:07 PM UTC (89a41a1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/encrypted-media/609/631787f...89a41a1.html" title="Last updated on Jun 9, 2026, 7:07 PM UTC (89a41a1)">Diff</a>

### DIFF
--- a/encrypted-media-respec.html
+++ b/encrypted-media-respec.html
@@ -3673,7 +3673,7 @@
                     <ol>
                       <li>
                         <p>
-                          If the <var>sanitized init data</var> is not supported by the
+                          If the <var>sanitized init data</var> is not supported by
                           <var>cdm</var>, reject <var>promise</var> with a {{NotSupportedError}}.
                         </p>
                       </li>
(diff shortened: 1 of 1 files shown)


encrypted-media-respec.html
xhwang-chromium: Should "the <var>cdm</var>" here also just be "<var>cdm</var>"? There are probably more cases like this.

chrisn: You're right, yes. I've fixed them all now.

xhwang-chromium (approved):

